PR #12, opened 2022-10-05
Ember Data model tweaks

Windvis:


app/models/generated-form.js
```diff
@@ -1,9 +1,9 @@
 import Model, { attr } from '@ember-data/model';
 
 export default class GeneratedFormModel extends Model {
-  @attr('string') created;
-  @attr('string') modified;
-  @attr('string') label;
-  @attr('string') comment;
-  @attr('string') ttlCode;
+  @attr('datetime') created;
```
Windvis, line 4: This will "break" some data that's already in the DB since they are not valid date strings. New data will have proper dates so it won't be an issue going forward.

PR #14, opened 2023-08-21
Fix: form not loading on initially opening the formbuilder

JonasVanHoof:
Sometimes when opening a form (frombuilder.edit) the form is not loaded into the builder or in the preview block. (only after a page refresh)
As for the test page I haven't noticed any problems locally with the form not loading on the user-test page. But felt like this could be togheter with the not loading of the form on the formbuilder edit page.


ZijneDwaasheid (commented):
Code looks alright to me.
It does what it needs to do: loading the form on opening. So no refresh needed anymore.

Not exactly something that should be addressed here, but I stumbled upon it when testing. When making a new blank form it displays the template form as it should, but does not save it to the store, because it only saves when clinking the "bewaar" button. That's why it appears to use the template, when you test it, it will be empty.

app/utils/forking-store-helper.js
```diff
@@ -0,0 +1,15 @@
+import { ForkingStore } from '@lblod/ember-submission-form-fields';
+
+export default class ForkingStoreHelper {
```
Windvis, line 3: Nitpick: Is there a benefit of using a class + static methods over exported util functions?

AFAIK, this setup has the downside that it can't be properly tree-shaken if we would switch to the new Embroider build system in the future.

app/controllers/formbuilder/edit.js
```diff
@@ -86,7 +85,11 @@ export default class FormbuilderEditController extends Controller {
       this.serializeSourceToTtl();
     });
 
-    this.firstRun = false;
+    if (isInitialRouteCall == true) {
```
Windvis, line 88:
```suggestion
    if (isInitialRouteCall) {
```

app/components/rdf-form-fields/property-group-selector.js
```diff
@@ -46,7 +46,13 @@ export default class PropertyGroupSelector extends InputFieldComponent {
         );
         return { subject: t.subject, label: label && label.value };
       });
-    this.options.sort(byLabel);
+
+    // There are option labels that are undefined in this list, I did not found it yet why or where these values are added but if they do not have a label they should not be visible in the frontend
+    const optionsWithALabel = allRawOptions.filter((_option) => {
```
Windvis, line 51: Nitpick: Why the `_undescored` variable names? 😄

app/controllers/formbuilder/edit.js
```diff
@@ -48,7 +50,7 @@ export default class FormbuilderEditController extends Controller {
     if (resetBuilder) {
       this.formChanged = true;
       this.builderStore.deregisterObserver();
-      this.builderStore = '';
+      this.builderStore = ForkingStoreHelper.getEmptyStoreValue();
```
Windvis, line 53: Nitpick / general feedback: I like the idea of renaming things more clearly so the code is readable, but I think it might be a bit overkill here. If I read the util name I expected that I would receive a new (empty) ForkingStore instance instead of an empty string, but maybe that's just me. 😄

I would probably just reset to `null` myself without a util.

Anyway, it seems that the "edit code flow" is broken at the moment though because of this line (also on master). Because of the async code happening after this, the page rerenders before a new ForkingStore instance is created and the form components receive an empty string instead of the instance which triggers an exception.

So this can probably be merged as-is and refactored in a separate PR when fixing the code editor flow.

Windvis (changes requested):
Other than 1 thing it's mostly some nitpicky stuff and questions 👍. Thanks!

app/components/rdf-form-fields/property-group-selector.js
JonasVanHoof: Thought:
This was done so it's clear that it is a parameter in a smaller scope than the main code. The variable option could be used in the main code and in the filter method. This also only visually because it would be using `this.option` but on the other side when passing arguments like in the refresh task here we can use one of these without the this keyword this `*refresh({ value, resetBuilder, isInitialRouteCall = false }) {`

Not sure how to handle those differences? Should we just use the parameter without the underscore?

app/utils/forking-store-helper.js
JonasVanHoof: Defining a class for it is Like the context for the method that will be called. For example you cannot create a method `createFromJson(json)` because you don't know what this will do. If you make a class in context of an Employee Controller for example you can call `Employee.createFromJson()` this example is actuallly not so great because you will not do this but I think this makes the reason behind it more clear?

Also I understand why we want to export these as functions and not as classes (after reading the `tree-shaken`)

Before I was writing symphony backend code so maybe it comes from there :)

=> will change it to functions in the future when I want to make a util

app/controllers/formbuilder/edit.js
```diff
@@ -96,12 +99,24 @@ export default class FormbuilderEditController extends Controller {
 
   @action
   serializeSourceToTtl() {
-    this.formChanged = true;
+    if (!RouterHelper.isCurrentlyOnRoute(this.router, 'formbuilder.edit')) {
```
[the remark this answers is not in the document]
JonasVanHoof: Created an [issue](https://github.com/lblod/frontend-form-builder/issues/17) for refactoring the `refresh task`. This is probably something that we should create as a Jira issue but im putting it here so we do not lose it out of sight

app/controllers/formbuilder/edit.js
```diff
@@ -35,10 +36,11 @@ export default class FormbuilderEditController extends Controller {
   graphs = GRAPHS;
   sourceNode = SOURCE_NODE;
 
-  @tracked firstRun = true;
+  @tracked isInitialDataLoaded = false;
 
   @task({ restartable: true })
-  *refresh({ value, resetBuilder }) {
+  *refresh({ value, resetBuilder, isInitialRouteCall = false }) {
```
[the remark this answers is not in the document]
JonasVanHoof: Renamed it to `formTtlCode`

app/controllers/formbuilder/edit.js
```diff
@@ -64,11 +66,8 @@ export default class FormbuilderEditController extends Controller {
       GRAPHS.formGraph
     );
 
-    let formRes = yield fetch(`/forms/form.ttl`);
-    let formTtl = yield formRes.text();
-
-    let metaRes = yield fetch(`/forms/meta.ttl`);
-    let metaTtl = yield metaRes.text();
+    const formTtl = yield this.getLocalFileContentAsText('/forms/form.ttl');
+    const metaTtl = yield this.getLocalFileContentAsText('/forms/meta.ttl');
```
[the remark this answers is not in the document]
JonasVanHoof: https://github.com/lblod/frontend-form-builder/issues/18

app/routes/formbuilder/edit.js
```diff
@@ -12,17 +12,41 @@ export default class FormbuilderEditRoute extends Route {
...
+      formTtlCode: this.getFormTtlCode(model),
+      resetBuilder: false,
+      isInitialRouteCall: true,
     });
   }
 
+  resetController(controller) {
+    if (
+      controller.builderStore &&
+      controller.builderStore.observers[controller.REGISTERED_FORM_TTL_CODE_KEY]
```
Windvis, line 31: I would consider `.observers` a private implementation detail of the class. `deregisterObserver` also doesn't fail if the key was already removed, so I would just always call it.

app/routes/formbuilder/edit.js
```diff
@@ -12,17 +12,41 @@ export default class FormbuilderEditRoute extends Route {
...
+    ) {
+      controller.builderStore.deregisterObserver();
```
Windvis, line 33: This still needs the `key` to properly remove the observer.

What I like to do is add a `reset` method to the controller and call that here. That way the controller can reset what it needs without the route having to know all that. That would also allow you to easily access the key constant you defined.

Windvis (changes requested):

app/routes/formbuilder/edit.js
JonasVanHoof: A good way of thinking thanks! [Changed](https://github.com/lblod/frontend-form-builder/pull/14/commits/955389ac216abe6c0090abc4d57fec37d5064658) it so the router just calls the controller method :ok_hand:

Windvis (approved):

PR #19, opened 2023-09-01
Feat/add validations to fields

JonasVanHoof:
Being able to show all the fields in the builded form and setting the validations that are possible on the field.


JonasVanHoof:
This PR is set `on hold` because we are trying to add the validations by generating them with the generators of the `ember-submission-form-field`

PR #21, opened 2023-09-19
Fix a slow redirect issue

Windvis:
Redirecting to the overview page from a very large form resulted in the browser freezing for several seconds. The underlying issue is observer system in the forking-store which gets used by the form fields addon. The latest patch fixes that issue upstream, so we update to that version to also fix the issue here.

With this PR leaving the details page of the "Big form" should no longer freeze the browser.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@ember/test-helpers": "^2.9.3",
         "@glimmer/component": "^1.1.2",
         "@glimmer/tracking": "^1.1.2",
-        "@lblod/ember-submission-form-fields": "^2.11.0",
+        "@lblod/ember-submission-form-fields": "^2.12.2",
         "broccoli-asset-rev": "^3.0.0",
         "codemirror": "^6.0.1",
         "codemirror-lang-turtle": "^0.0.2",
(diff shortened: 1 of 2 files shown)


JonasVanHoof:
Sorry not getting any emails from this :thinking: 

Pulled master and created a form with 50 fields (copied the ttl from the slow form) and this will not freeze the application when going back to the overview of the forms GJ :)

Windvis:
It seems the redirect issue wasn't the only slowness issue. I still see browser freezes when clicking and leaving form fields 🤔 That might be harder to fix.

JonasVanHoof:
> It seems the redirect issue wasn't the only slowness issue. I still see browser freezes when clicking and leaving form fields 🤔 That might be harder to fix.
This is only the case when changing the values of fields in a big form. Does this mean we should look into a more optimized way of showing and updating the form. eg split it into pieces and check which one should be updated?

Windvis:
For me it also happens on blank forms, for example this one: https://dev.form-builder.lblod.info/formbuilder/650383E498CA4C0008000013/edit but only with my local frontend, so I might have broken things with other changes 😬. Or it's only the dev build that has the issue.

To reproduce I simply click in the "section title" field, edit it and click outside. After that Firefox freezes and it asks me to kill the script 😄.

I'll investigate some more after lunch.

Windvis:
I pinned it down a change in the forking store: https://github.com/redpencilio/forking-store/pull/24

Whoops 😬.

PR #22, opened 2023-09-19
feat: generate a form where you can add validations to the field

JonasVanHoof:
Issue: [SBF-3](https://binnenland.atlassian.net/browse/SFB-3)


JonasVanHoof:
| Done    | Todo                                                                                        | Comment                                                                                                                 |
| ------- | ------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| &check; | Validation is added correctly                                                               |                                                                                                                         |
| &check; | Validation is removed correctly                                                             |                                                                                                                         |
| &cross; | validation can be edited correctly                                                          | When assigning a required constrained clicking in the preview field and going out it adds the rcross that this is wrong |
| &cross; | You cannot add duplicate validations                                                        | You can add dupklicate validations. They are shown twice in the preview                                                 |
| &cross; | You can only add the validation when it is possible to add to the field (canHaveValidation) | Not implememted                                                                                                         |
| &cross; | You can only add the validation when it is possible to add to the field (canHaveValidation) | Not implememted                                                                                                         |

JonasVanHoof:
@Windvis this PR should go about the `requiredConstrained` but the `maxLength` is also in here as we wahere not able to add the validations and show the fields when no conditional field was visible, so you will also find code for the maxLength validation.

The issue when selecting the maxLength is now that the red border is shown. This is because the grouping is still `Bag` instead of `MatchEvery`. After this is merged I can look at updating the grouping types for each validation in the form to the ones defined in the `validations/meta.ttl`

JonasVanHoof:
Also the idea was to be able to pass enum values to t `a validate` method, the will than check all passed validations. But as in our usecase it should always validate every validation in the form so this can wait.. Will be improving the `areValidationsInGraphValidated` while adding the validator checks for upcoming validations.

.gitignore
```diff
@@ -2,6 +2,7 @@
 
 # IDE
 .idea
+.vscode
```
Windvis, line 5: nitpick: Git has an option to use a ["global gitignore"](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file). That is where I add things specific to my development environment. I wouldn't add it here.

app/components/add-validations-to-form.js
```diff
@@ -0,0 +1,74 @@
...
+import { ForkingStore } from '@lblod/ember-submission-form-fields';
+import { getLocalFileContentAsText } from '../utils/get-local-file-content';
+import { FORM, RDF } from '../utils/rdflib';
+import { areValidationsInGraphValidated } from '../utils/validation-shape-validators';
+
+export default class AddValidationsToFormComponent extends Component {
+  @tracked builderStore;
+  @tracked builderForm;
+  @tracked formTtlCode;
+  @tracked showRdfForm = false;
```
Windvis, line 13: Nitpick: it's usually better to use derived state if possible. Either by moving the logic to an ember-concurrency task (which has a running state) or by using a getter that returns true if builderStore and builderForm aren't falsy.

app/components/add-validations-to-form.js
```diff
@@ -0,0 +1,74 @@
...
+
+  constructor() {
+    super(...arguments);
+
+    if (!this.args.formTtlCode || this.args.formTtlCode == '') {
+      throw `Cannot add validations to an empty form`;
+    }
+
+    this.formTtlCode = this.args.formTtlCode;
+    this.sourceNode = this.args.sourceNode;
```
Windvis, line 29: Is this needed? I would use the args if there is no specific reason.

app/components/add-validations-to-form.js
```diff
@@ -0,0 +1,74 @@
...
+
+    this.createBuilderStore().then((builderStore) => {
```
Windvis, line 31: I would move this to an async method or an ember-concurrency task and simply start it in the constructor.

app/components/add-validations-to-form.js
```diff
@@ -0,0 +1,74 @@
...
+export default class AddValidationsToFormComponent extends Component {
+  @tracked builderStore;
+  @tracked builderForm;
+  @tracked formTtlCode;
+  @tracked showRdfForm = false;
+  @tracked sourceNode;
+
+  graphs = GRAPHS;
+  form_ttl_path = '/forms/validation/form.ttl';
+  meta_ttl_path = '/forms/validation/meta.ttl';
```
Windvis, line 18: nitpick: What's up with the snake_case naming? 😄

app/components/add-validations-to-form.js
```diff
@@ -0,0 +1,74 @@
...
+      await getLocalFileContentAsText(this.meta_ttl_path),
+      GRAPHS.metaGraph,
+      'text/turtle'
+    );
+    builderStore.parse(this.formTtlCode, GRAPHS.sourceGraph, 'text/turtle');
+
+    return builderStore;
+  }
+
+  createBuilderForm(builderStore) {
```
Windvis, line 66: Nitpick: the name is a bit confusing since nothing is created, really. It's retrieved from the graph.

app/components/add-validations-to-form.hbs
```diff
@@ -0,0 +1,13 @@
+{{#if this.showRdfForm}}
+  <RdfForm
+    @groupClass="au-o-grid__item"
+    @form={{this.builderForm}}
+    @graphs={{this.graphs}}
+    @formStore={{this.builderStore}}
+    @forceShowErrors={{true}}
```
Windvis, line 7: Does this need to be set to `true`? This forces the error state for fields even if the user didn't touch them yet.

app/styles/_shame.scss
```diff
@@ -50,4 +50,8 @@ code {
 
 .au-o-grid {
   display: block;
-}
\ No newline at end of file
+}
+
+.playground__edit {
```
Windvis, line 55: I guess this belongs in the `c-playground.scss` file?

app/styles/project/_c-playground.scss
```diff
@@ -40,3 +40,11 @@
   overflow-x: hidden;
   background-color: $au-gray-100;
 }
+
+.playground__tab {
```
Windvis, line 44: Nitpick, I would rename the class to `.playground-tab`. It's not a child element of the `playground` block, so it feels weird to use the element notation here.

app/templates/formbuilder/edit.hbs
```diff
@@ -11,10 +11,18 @@
 
 <main id="main" class="au-c-main-container">
   <div class="au-c-main-container__content">
+    <div class="playground__tab">
+      Builder
```
Windvis, line 15: The toggle is a bit weird since you can only click the "validation" label to toggle it. Not sure what the fix for that would be though so something for the UX team in one of the future sprints.

JonasVanHoof:
If this can be merged we can check if we want to add the validation concept scheme selector as followed https://github.com/lblod/frontend-form-builder/pull/32

public/forms/validation/form.ttl
```diff
@@ -0,0 +1,159 @@
...
+# Form
+##########################################################
+ext:form a form:Form, form:TopLevelForm ;
+  form:includes ext:formNodesL .
+
+##########################################################
+#  Property-groups
+##########################################################
+ext:formFieldPg a form:PropertyGroup;
+  sh:name "Field" ;
```
Windvis, line 23: Hmm, we need to tweak the property group setup in a future PR so that it doesn't show this title everywhere

public/forms/validation/form.ttl
```diff
@@ -0,0 +1,159 @@
...
+    #  Max-lenght field
+    ##########################################################
+    ext:validationMaxLengthF a form:Field ;
+      sh:name "Max characters" ;
+      sh:order 101 ;
+      sh:path form:max ;
+      form:validations 
+        [ a form:RequiredConstraint ;
+            form:grouping form:Bag ;
+            sh:path rdf:type ;
```
Windvis, line 155:
```suggestion
            sh:path form:max;
```

It doesn't trigger the validation since the type is always set.

public/forms/validation/form.ttl
```diff
@@ -0,0 +1,159 @@
...
+
+##########################################################
+#  Validation type conditions
+##########################################################
+fields:maxLengthValidationCFG a form:ConditionalFieldGroup ;
+  form:conditions
+    [ a form:SingleCodelistValue ;
+        form:grouping form:Bag ;
+        sh:path rdf:type ;
+        form:conceptScheme concept:dde3d2a3-e848-47ea-ba44-0f2e565f04ab ;
```
Windvis, line 127: I would give the concept a more descriptive uri instead of a UUID, or is this something that will be used outside of the form-builder as well?

.gitignore
JonasVanHoof: Didn't know that NICE

app/components/add-validations-to-form.js
JonasVanHoof: It's just to make it clear that these are not variables that.. yeah I can change it so just camelCasing :sweat_smile:

Windvis: Yea, in that case I would make them const variables in the module scope, or simply use the strings in the function directly 😄.

app/templates/formbuilder/edit.hbs
JonasVanHoof: Indeed putting in time and effort now to make it work is a bit of a waste. On Monday I also sat down with Marie and there we said that we are going to take a look on how we are going to organise everything within the form builder design wise & placements.

public/forms/validation/form.ttl
JonasVanHoof: https://github.com/lblod/frontend-form-builder/issues/33

public/forms/validation/form.ttl
JonasVanHoof: Indeed I now see the difference :D
image

public/forms/validation/form.ttl
JonasVanHoof: I could change it in the meta.ttl file as I only found [results on this ID](https://github.com/search?q=org%3Alblod%20dde3d2a3-e848-47ea-ba44-0f2e565f04ab&type=code) in the `form-builder` itself.

Does this also mean that we will be having  to create a migration for these as they are added to the database in the [backend](https://github.com/lblod/app-form-builder/blob/7e74170f0a26932a5fc58667919a865eb70dc642/config/migrations/20210305123324-add-validations.sparql#L11)?

app/components/add-validations-to-form.js
JonasVanHoof: Directly created the builderForm in the callback method when creating the builderStore

app/components/add-validations-to-form.js
JonasVanHoof: Created a `getter` for it + renamed it to `canShowRdfForm`

public/forms/validation/form.ttl
Windvis: Maybe, I'm not sure why we inserted these in the DB actually. Don't we only need them in the frontend?

JonasVanHoof: Than this was a good question :D It actually confused me a bit because when changing the objects in the `meta.ttl` (for example: the `form:required` in meta.ttl  now was a full uri before) file it actually just worked in the frontend so that made me think that it is indeed not needed that it is stored in the database. 

Another thing thing is when starting with the validation form @aatauil  got me quering the database and putting that ttl result in the valdiations meta ttl file. This again confirms that we do not need to write migrations for it?

Ill change the id to something more descriptive :)

app/styles/_shame.scss
JonasVanHoof: The buttons disappeared when adding more fields but added a margin now for the `au-o-box` in the playground that fixes it for the builder form and the validation form 
```css
.au-o-box {
  margin-bottom: 4rem;
}

Comment (Windvis) [reply app/styles/_shame.scss]: Hmmmmmmmmm 😄, I wouldn't extend Appuniversum classes like that since it will also affect other cases (or future cases we might add). We probably can fix it by restructuring the html setup, but as a quick fix I would use one of the margin spacing utils: https://appuniversum.github.io/ember-appuniversum/?path=/story/utilities-spacing--page

File: public/forms/validation/form.ttl (hunk 13 above)
Comment (JonasVanHoof) [reply public/forms/validation/form.ttl]: Changed it to `http://lblod.data.gift/concept-schemes/possibleValidations`

File: app/styles/_shame.scss (hunk 8 above)
Comment (JonasVanHoof) [reply app/styles/_shame.scss]: Woeps haha

It is fixed with the spacing style on the builder form but not yet fully happy about it in the validation form. But it "workable"

File: app/controllers/formbuilder/edit.js [hunk 14]
```diff
@@ -36,6 +37,33 @@ export default class FormbuilderEditController extends Controller {
 
   @tracked isInitialDataLoaded = false;
 
+  @tracked isShowBuilder = true;
+  @tracked fieldsInForm = [];
+  formValidationService = null;
```
JonasVanHoof, line 42: Also try to use the derived state here as [mentioned](https://github.com/lblod/frontend-form-builder/pull/22#discussion_r1344071910)

Windvis (approved):

app/controllers/formbuilder/edit.js
Windvis: This doesn't seem to be used (yet)? Does this block the merging of this PR?

JonasVanHoof: Definitly not :D

JonasVanHoof:
@Windvis im not merging it yet because I saw that not all the data is shown in the fields when switching between builder and validations tab. Also when changing somehting in the validation form there is not possibily to `Bewaar` the current changes

Windvis:
Whoops, my bad. I thought you meant I could merge it 😬

Windvis:
> Also when changing somehting in the validation form there is not possibily to Bewaar the current changes

I guess we also need to modify the `formChanged` value in the `onUpdateValidations` handler?

PR #26, opened 2023-09-19
Resolve Appuniversum deprecations

Windvis:
The remaining deprecations come from the form-fields addon.


app/templates/application.hbs
```diff
@@ -1,11 +1,13 @@
 {{page-title "Semantic form builder"}}
 
+<AuModalContainer />
```
Windvis, line 3: We can't move it below the outlet since some routes add a `AuModal` which isn't conditionally rendered and that throws an error if the container isn't found.

PR #28, opened 2023-09-20
Enable the Embroider build system

Windvis:
Embroider is the new build system for Ember that uses Webpack behind the scenes. This should result in faster (re)builds and smaller bundles without any real code changes.

More information: https://github.com/embroider-build/embroider

(You might need to install Watchman if you get file watching issues during the build)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ You will need the following things properly installed on your computer.
 * [Node.js](https://nodejs.org/) (with npm)
 * [Ember CLI](https://cli.emberjs.com/release/)
 * [Google Chrome](https://google.com/chrome/)
+* [Watchman](https://facebook.github.io/watchman/) (might be needed on some OSs if you run into file watching errors)
 
 ## Installation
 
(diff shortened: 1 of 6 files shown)


JonasVanHoof:
@Windvis 
Checked out the branch.
Overall application speed is improved NICE. Didn't have any issues with the watcher.

One thing I noticed that their are a lot more deprecation's showing in the console now. It's possible to disable those in browser when developing and checking at the end if there are any new deprecations by implementing the new stuff. Not sure if we can do something or should do something about this spamming the console?

JonasVanHoof (approved):
Checked out the branch.
Overall application speed is improved NICE. Didn't have any issues with the watcher.

One thing I noticed that their are a lot more deprecation's showing in the console now. It's possible to disable those in browser when developing and checking at the end if there are any new deprecations by implementing the new stuff. Not sure if we can do something or should do something about this spamming the console?

Windvis:
I think those deprecations come from Appuniversum and are the result of the Appuniversum update in #25 (and I resolved the ones I could in #26). The remaining ones come from the ember-submission-form-fields addon but I haven't gotten to it yet to update it.

It is possible to hide specific deprecations with this: https://api.emberjs.com/ember/5.2/functions/@ember%2Fdebug/registerDeprecationHandler

I didn't think it was worth adding such a setup here since I was planning on resolving the upstream issue soon anyways. You can definitely add it in your app.js file if you want to silence them for now though.

PR #31, opened 2023-09-28
Add a conditional "max length" field

Windvis:
This tweaks the conditional field setup so it works for the "max length" field.

### DIFF
--- a/public/forms/validation/form.ttl
+++ b/public/forms/validation/form.ttl
@@ -26,7 +26,6 @@ ext:formFieldValidations a form:PropertyGroup;
   sh:name "Validation" ;
   sh:order 2 .
 
-
 ##########################################################
 #  Scopes:
 ##########################################################
@@ -61,9 +60,7 @@ ext:formNodesL a form:Listing;
 ##########################################################
 ext:formValidationItem a form:SubForm;
   form:includes ext:formNodesNameF;
-  form:includes ext:validationResultMessageF;
-  form:includes ext:formValidationL;
-  form:hasFieldGroup ext:formFieldValidations .
+  form:includes ext:formValidationL.
 
 ##########################################################
 #  Validation Listing:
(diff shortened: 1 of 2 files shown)


public/forms/validation/form.ttl
```diff
@@ -79,9 +76,12 @@ ext:formValidationL a form:Listing;
   sh:group ext:formFieldPg;
   sh:order 3 .
 
-ext:formValidationValueItem a form:SubForm;
-  form:includes fields:validatonSelectorF;
-  form:hasFieldGroup ext:formFieldValidations .
+ext:formValidationValueItem a form:SubForm ;
+  form:hasFieldGroup ext:validationsFG ;
+  form:includes ext:validationResultMessageF .
```
Windvis, line 81: It seems that we _need at least 1 form:includes_ in a subform for it to render.. this is most likely a bug in the addon. It works out here since I think we _always_ want the error message field?

public/forms/validation/form.ttl
```diff
@@ -101,50 +101,61 @@ fields:validatonSelectorF a form:Field ;
...
+
+##########################################################
+#  Error message field
+##########################################################
+ext:validationResultMessageF a form:Field ;
+  sh:name "Error message" ;
+  sh:order 100 ; # I think we always want to show this below the conditional fields?
+  sh:path sh:resultMessage ;
+  form:displayType displayTypes:defaultInput;
+  sh:group ext:formFieldPg .
```
Windvis, line 115: I _think_ they all need to be the same PG, I think they didn't render when I used the second one, but maybe something else was wrong.

public/forms/validation/meta.ttl
```diff
@@ -116,12 +116,14 @@ ns6:c09902d3-9421-49e9-a2e6-d3ac42abbca2	rdf:type	rdfs:Class ,
 	form:grouping	"MatchEvery"@en ;
 	sh:resultMessage	"Geef een geldig telefoonnummer in."@nl ;
 	ext:customParameter	"defaultCountry"@en .
-<http://lblod.data.gift/concepts/60873518-5423-4121-ba6a-9635b18242a0>	rdf:type	rdfs:Class ,
```
Windvis, line 119: These need to be the constraint class uris I think, I only adjusted it for the MaxLength one for now.

public/forms/validation/form.ttl
Windvis: Hmm, no that doesn't seem to be the case, so this can be improved since it now shows "Field" everywhere 😄

PR #34, opened 2023-10-04
Fix: data is not loaded in builder form when switching between tabs

JonasVanHoof:
This is already fixing the loading issue

### DIFF
--- a/app/components/add-validations-to-form.js
+++ b/app/components/add-validations-to-form.js
@@ -49,6 +49,12 @@ export default class AddValidationsToFormComponent extends Component {
     }, REGISTERED_FORM_TTL_CODE_KEY);
   }
 
+  willDestroy() {
+    super.willDestroy(...arguments);
+
+    this.builderStore.deregisterObserver(REGISTERED_FORM_TTL_CODE_KEY);
+  }
+
   serializeToTtlCode(builderStore) {
     const sourceTtl = builderStore.serializeDataMergedGraph(
       this.graphs.sourceGraph


app/components/add-validations-to-form.js
```diff
@@ -46,10 +49,19 @@ export default class AddValidationsToFormComponent extends Component {
 
     this.builderStore.registerObserver(() => {
       this.serializeToTtlCode(this.builderStore);
-    }, REGISTERED_FORM_TTL_CODE_KEY);
+    }, OBSERVER_KEY.validationForm);
   }
 
   serializeToTtlCode(builderStore) {
+    if (this.args.isComponentNotShown) {
```
Windvis, line 56: Wait, isn't the real issue that we don't deregister when the component is being destroyed (when switching tabs) I think using the [willDestroy](https://api.emberjs.com/ember/release/modules/@glimmer%2Fcomponent#willdestroy) hook would be a better place to do that.

If the early return here is still needed after that I think you can also just check the `this.isDestroying` or `this.isDestroyed` properties.

app/utils/deregister-forking-store-observable.js
Windvis, line 1: Sorry, but this all feels a bit overkill to me. I don't see real value in moving the observer keys in a shared file, and the deregister logic feels unneeded as well. We shouldn't need to check if we're dealing with a forking store since that should always be the case at that point (unless something is wrong with the code flow).

This feels a bit like a premature abstraction or trying to be overly safe. I would just try to keep it simple as much as possible.

app/controllers/formbuilder/edit.js
```diff
@@ -106,7 +109,7 @@ export default class FormbuilderEditController extends Controller {
 
     this.builderStore.registerObserver(() => {
       this.serializeSourceToTtl();
-    }, REGISTERED_FORM_TTL_CODE_KEY);
+    }, OBSERVER_KEY.builderForm);
```
Windvis, line 112: Note: the latest forking-store release [also properly supports using the observer function as the key](https://github.com/redpencilio/forking-store/pull/22), so we can remove the string key if you want.

app/components/add-validations-to-form.js
```diff
@@ -8,12 +8,15 @@ import { getLocalFileContentAsText } from '../utils/get-local-file-content';
...
-const REGISTERED_FORM_TTL_CODE_KEY = 'validationFormTtlCode';
+import {
+  OBSERVER_KEY,
+  deregisterFromForkingStoreObservable,
+} from '../utils/deregister-forking-store-observable';
 
 export default class AddValidationsToFormComponent extends Component {
   @tracked builderStore;
   @tracked builderForm;
+  @tracked isComponentNotShown;
```
Windvis, line 19: Note: not sure if this was an oversight, or intentionally added to match the arg. but tracked properties are unrelated to the arguments so this wasn't needed.

Windvis (changes requested):

app/components/add-validations-to-form.js
JonasVanHoof: I was looking for a hook but only found [these](https://guides.emberjs.com/v2.14.0/components/the-component-lifecycle/) and they where never called when testing this with console logs.. this willwillDestroy hook does the same and is indeed better

Windvis: Yea, Ember 2.14 is _very_ old 😄. Those are the docs for the old "ember component" which are different from the Glimmer components we are using now.

app/components/add-validations-to-form.js
JonasVanHoof: Thats good to know because it felt like this was always needed..

app/components/add-validations-to-form.js
JonasVanHoof: Damned always check the version indeed and than the type of component used is also important ..

app/controllers/formbuilder/edit.js
JonasVanHoof: Can we bump the forking store version in the ember-submission-form-fields to  `2.1.1` for this?

Windvis: I think we already have that version: https://github.com/lblod/frontend-form-builder/blob/656b041d1c51055f1507d9f054f22ef94fc0611a/package-lock.json#L38642

image

JonasVanHoof: hm hm hm Im going to check my setup real quick 
image

Windvis (approved):

PR #35, opened 2023-10-04
Add listing item reordering

Windvis:
This uses the new reordering feature to the builder form listings.

I didn't add it to the "validations" listing since the order there doesn't really matter _I think_?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@embroider/webpack": "^3.0.0",
         "@glimmer/component": "^1.1.2",
         "@glimmer/tracking": "^1.1.2",
-        "@lblod/ember-submission-form-fields": "^2.13.0",
+        "@lblod/ember-submission-form-fields": "^2.14.1",
         "broccoli-asset-rev": "^3.0.0",
         "codemirror": "^6.0.1",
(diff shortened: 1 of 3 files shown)


JonasVanHoof:
> This uses the new reordering feature for the builder form listings.
> 
> I didn't add it to the "validations" listing since the order there doesn't really matter _I think_?

Atm this will not be needed thanks for checking!

JonasVanHoof (approved):

Windvis:
Hmmmm, I noticed something strange with the order field now. After changing the order with the new buttons, the preview is updated and it shows the correct order, but the value in the order field isn't updated.

The input field doesn't have an observer so it doesn't update the visible value even if the graph was updated. Should we just remove the order input field and depend on the order in the builder form?

Windvis:
Aaaand I noticed a regression in the addon 😬 whoops.

Windvis:
> Edit: I removed the order field for now. It has the disadvantage that you can't order things by hand but I don't think we want users to do that anyways. This could become an issue because of the listings though. They are fields in the same property group, but their ordering count is separate from the other fields because they are created in a different listing 🤔.

It seems it wasn't possible to modify the order of the listings compared to the normal fields in the previous version either, so this didn't break that.

PR #36, opened 2023-10-04
Feat: create validation concept scheme selector

JonasVanHoof:
Issue: [SFB-3](https://binnenland.atlassian.net/jira/software/c/projects/SFB/issues/SFB-3)

This selector will only show the validations that are possible for that display type (field).
Each field has a predicate `canHaveValidation` and the options of the dropdown are based on that

### DIFF
--- a/app/components/rdf-form-fields/validation-concept-scheme-selector.hbs
+++ b/app/components/rdf-form-fields/validation-concept-scheme-selector.hbs
@@ -1,0 +1,34 @@
+{{#unless @inTable}}
+  <AuLabel
+    @error={{this.hasErrors}}
+    @required={{this.isRequired}}
+    for={{this.inputId}}
+  >
+    {{@field.label}}
+  </AuLabel>
+{{/unless}}
+<div class={{if this.hasErrors "ember-power-select--error"}}>
+  <PowerSelect
+    @triggerId={{this.inputId}}
+    @searchField="label"
+    @searchEnabled={{this.searchEnabled}}
+    @selected={{this.selected}}
+    @options={{this.options}}
+    @onClose={{fn (mut this.hasBeenFocused) true}}
+    @onChange={{this.updateSelection}}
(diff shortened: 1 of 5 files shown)


JonasVanHoof:
@Windvis  do we want to have a dependency on the `concept-scheme-selector` from the `ember-submission-form-fields` ? Otherwise I think it is possible to use that component and use the `validation-concept-scheme-selector` as a wrapper around that component

Windvis:
Hard to say without looking at the code. What do you mean with a wrapper?

I wouldn't go the `class ValidationConceptSchemeSelector extends ConceptSchemeSelector {}` route since I want to move away from that.

If you want to use it inside the .hbs file I think that's fine 👍 (if it's possible to make it work that way, not sure if the component is flexible enough).

JonasVanHoof:
> Hard to say without looking at the code. What do you mean with a wrapper?
> 
> I wouldn't go the `class ValidationConceptSchemeSelector extends ConceptSchemeSelector {}` route since I want to move away from that.
> 
> If you want to use it inside the .hbs file I think that's fine 👍 (if it's possible to make it work that way, not sure if the component is flexible enough).

Yes the idea was that i use the component in my hbs file and than update evrything that is needed so that we do have almost the same files in two projects.

If I can't make it work the current version will be it ;p

JonasVanHoof:
@Windvis  Can't seem to figure out atm on how it should work with the ember-submission-form-fields component. Would like to focus on the saving, not being able to add duplicates and that when you create a field with another displaytype than default input in the builder that it shows as a normal defaultInputField in the validation form..

Created an [issue](https://github.com/lblod/frontend-form-builder/issues/37)  (enhancement) for updating the validation-concept-scheem-selector here in the form-builder frontend

Example:
| Builder  | Validation   |
|----------|--------------|
| DateTime | DefaultInput |
| TextArea | DefaultInput |

Issue #17, opened 2023-08-23
Cleanup the refresh task in the FormbuilderEditController


JonasVanHoof:
https://github.com/lblod/frontend-form-builder/pull/14#discussion_r1301625781

PR #38, opened 2023-10-05
Fix: saving the form data

JonasVanHoof:
Everytime you go to the builder tab the refresh method is called with the `isInitialRouteCall` parameter on `true` this makes it that the `formChanged` variable is set to false as when initially lodaing the data there is no way that you already changed something in the form. 

I added a saved state of the variable `formChanged` to overwrite this but it feels that im hAcKinG more than we should @Windvis

### DIFF
--- a/app/controllers/formbuilder/edit.js
+++ b/app/controllers/formbuilder/edit.js
@@ -48,9 +48,7 @@ export default class FormbuilderEditController extends Controller {
         isInitialRouteCall: true,
       });
 
-      if (isFormChangedSavedState) {
-        this.setFormChanged(true);
-      }
+      this.setFormChanged(isFormChangedSavedState);
     } else {
       this.deregisterFromObservable();
     }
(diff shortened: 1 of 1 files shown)


Windvis:
Any idea why the`isInitialRouteCall` needs to be set to true here in the first place?

Windvis:
Anyway, it feels like we should prioritize https://github.com/lblod/frontend-form-builder/issues/17 😄 Things are hard to follow and it doesn't get easier with the extra form.

JonasVanHoof:
> Any idea why the`isInitialRouteCall` needs to be set to true here in the first place?

This is done only to set these variable as when you initially load the form you could never have changed something in the builder (before it showed that the state is changed because it had already serialized the store graphs)
```js
    if (isInitialRouteCall) {
      this.setFormChanged(false);
      this.isAddingValidationToForm = false;
      this.isShowBuilder = true;
    }

PR #41, opened 2023-10-16
Feat: max length validation is shown without the error border

JonasVanHoof:
Issue: [SFB-4](https://binnenland.atlassian.net/browse/SFB-4)


JonasVanHoof:
image

JonasVanHoof:
It generated the grouping types of the validations as literals.. not sure if there is another way of fixing this

Windvis (commented):
I'm a bit confused. If I understand it correctly this will check the form "grouping" values by comparing it to an expected one in the meta graph and replace it with the expected one if it doesn't match? How do we get in that invalid state? Does the selector component add the wrong value?

app/utils/get-grouping-type-for-validation.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!groupingType) {
+    return FORM('Bag');
+  }
+
+  return getNamedNodeGroupingType(groupingType);
+}
+
+function getNamedNodeGroupingType(groupingType) {
+  if (isLiteral(groupingType)) {
```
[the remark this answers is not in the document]
JonasVanHoof: Facepalm didn't link those two together for some reason.. Updated the `meta.ttl` file with NamedNode values for the `form:grouping`

Windvis:
> as we cannot dynamically set this value depending on the validation type if im correct

Can't we do this in the custom selector component? Or did that not work? If we need to do it this way I would add a comment somewhere that explains why it's needed. It looks a bit strange if you don't know this background context.

JonasVanHoof:
> > as we cannot dynamically set this value depending on the validation type if im correct
> 
> Can't we do this in the custom selector component? Or did that not work? If we need to do it this way I would add a comment somewhere that explains why it's needed. It looks a bit strange if you don't know this background context.

Hm good thinking! I just checked it but the selector is only for the type of the valdiation and im not seeing on how to pass the grouping type of the validation with that selector to the generated validation node.. any idea?

Otherwise it is indeed nice to add a comment here so people know this is because we cannot dynamically update the grouping type depending on the validation type in the form.ttl file.

Windvis:
> Hm good thinking! I just checked it but the selector is only for the type of the valdiation and im not seeing on how to pass the grouping type of the validation with that selector to the generated validation node.. any idea?

I'm not sure, but can't we simply do the 'expected type' lookup when the user selects a validation type and set both the type and grouping predicates? Instead of only setting the type? It works in my head at least 😄.

JonasVanHoof:
> > Hm good thinking! I just checked it but the selector is only for the type of the valdiation and im not seeing on how to pass the grouping type of the validation with that selector to the generated validation node.. any idea?
> 
> I'm not sure, but can't we simply do the 'expected type' lookup when the user selects a validation type and set both the type and grouping predicates? Instead of only setting the type? It works in my head at least 😄.

Alright found a solution and indeed it's the way that was in your head? Only thinking that i move the method for adding to a separate file, what do you think?

public/forms/validation/form.ttl
```diff
@@ -85,7 +85,7 @@ ext:validationsFG a form:FieldGroup ;
 fields:validatonSelectorF a form:Field ;
   sh:name "Validation type" ;
   sh:order 2 ;
-  sh:path rdf:type ;
+  sh:path rdf:type, form:grouping ;
```
Windvis, line 88: Oh, I see what you were asking before. Since we're in custom component space we can just make assumptions in the component itself, so this wasn't strictly needed. But fine to keep if it works 👍.

app/components/rdf-form-fields/validation-concept-scheme-selector.js
```diff
@@ -147,7 +149,38 @@ export default class ValidationConceptSchemeSelectorComponent extends InputField
...
+      RDF('type'),
+      validationType,
+      graphs.sourceGraph
+    );
+    const groupingType = getGroupingTypeForValidation(
+      validationType,
+      store,
+      graphs.metaGraph
+    );
+    store.addAll([
```
Windvis, line 177: I think we also need to remove the potential "old" value here?

app/components/rdf-form-fields/validation-concept-scheme-selector.js
```diff
@@ -147,7 +149,38 @@ export default class ValidationConceptSchemeSelectorComponent extends InputField
       updateSimpleFormValue(this.storeOptions, option.subject);
```
Windvis, line 149: I would remove this and update both the type and grouping at the same time (addAll + removeStatements call). That will trigger the observers "only" twice (until I add a new method to the forking store to do both inserts and deletes in a single method call).

public/forms/validation/form.ttl
JonasVanHoof: Perfect! Just found out that it works like this :D

app/components/rdf-form-fields/validation-concept-scheme-selector.js
JonasVanHoof: Adding the rdf type and the form grouping in one addAll([]) call

app/components/rdf-form-fields/validation-concept-scheme-selector.js
JonasVanHoof: There are no remains in the code to clear up

Windvis: Where do those old values get removed then?

JonasVanHoof: @Windvis  I'm in a small conflict on what to do now.. I know I have to remove the grouping and the RDF type in this component. This assigning and updating was done by using the sh:path that is passed on to this component. Are we just removing this logic with the `path`as we know it needs to set and update the rdf- and the grouping type?

On the one side we lean to making this component ours by doing this.. instead of the method `updateSimpleValue` and on the other side we are still using this `path` thingy?
```js
      this.storeOptions.store.addAll([
        new Statement(
          this.storeOptions.sourceNode,
          FORM('grouping'),
          groupingType,
          this.storeOptions.sourceGraph
        ),
        new Statement(
          this.storeOptions.sourceNode,
          RDF('type'),
          option.subject,
          this.storeOptions.sourceGraph
        ),
      ]);
```

What should be a good solution? If we are not going to use the `sh:path` we can remove this.. but if we are still going to use it we will be adding/removing/updating the validation `type` with this code and the grouping in another way what makes it hard to know what is going on..
```js
const matches = triplesForPath(this.storeOptions, true).values;
      this.selected = this.options.find((opt) =>
        matches.find((m) => m.equals(opt.subject))
      );
```

JonasVanHoof: We agreed on removing the `path` stuff in the component. This is updated in the code to adding and removing the items by the predicate key (`rdf:type` & `form:grouping`)

PR #42, opened 2023-10-17
Refactor the "refresh" task

Windvis:
This refactors the setup code so it's more flexible and easier to follow.

It's still rough and not very well tested, so most likely incomplete,  but I won't be able to finish for a while..

### DIFF
--- a/app/components/code-edit-modal.js
+++ b/app/components/code-edit-modal.js
@@ -1,0 +1,9 @@
+import Component from '@glimmer/component';
+import { restartableTask, timeout } from 'ember-concurrency';
+
+export default class CodeEditModal extends Component {
+  handleCodeChange = restartableTask(async (newCode) => {
+    await timeout(500);
+    this.args.onCodeChange?.(newCode);
+  });
+}
--- a/app/controllers/formbuilder/edit.js
+++ b/app/controllers/formbuilder/edit.js
@@ -2,12 +2,12 @@ import Controller from '@ember/controller';
 
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
-import { task, timeout } from 'ember-concurrency';
+import { restartableTask, timeout } from 'ember-concurrency';
 import { inject as service } from '@ember/service';
 import { ForkingStore } from '@lblod/ember-submission-form-fields';
 import { sym as RDFNode } from 'rdflib';
 import { FORM, RDF } from '../../utils/rdflib';
-import { getLocalFileContentAsText } from '../../utils/get-local-file-content';
+import basicFormTemplate from '../../utils/basic-form-template';
 
 export const GRAPHS = {
   formGraph: new RDFNode('http://data.lblod.info/form'),
@@ -18,10 +18,9 @@ export const GRAPHS = {
 const SOURCE_NODE = new RDFNode('http://frontend.poc.form.builder/sourcenode');
 
 export default class FormbuilderEditController extends Controller {
-  @service('meta-data-extractor') meta;
   @service store;
 
-  @tracked code;
+  @tracked formCode;
 
   @tracked previewStore;
   @tracked previewForm;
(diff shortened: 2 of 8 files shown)


app/components/code-edit-modal.js
```diff
@@ -0,0 +1,9 @@
+import Component from '@glimmer/component';
+import { restartableTask, timeout } from 'ember-concurrency';
+
+export default class CodeEditModal extends Component {
+  handleCodeChange = restartableTask(async (newCode) => {
+    await timeout(500);
```
Windvis, line 6: Moved the long timeout here. We still want to prevent multiple events while the user is still typing.

app/templates/index.hbs
```diff
@@ -31,16 +31,16 @@
...
-            <td class="au-u-visible-medium-up">{{row.comment}}</td>
-            <td class="au-u-visible-medium-up">{{format-date row.created}}</td>
-            <td class="au-u-visible-medium-up">{{format-date-time row.modified}}</td>
+          <c.body as |generatedForm|>
+            <td class="au-u-visible-medium-up">{{generatedForm.label}}</td>
+            <td class="au-u-visible-medium-up">{{generatedForm.comment}}</td>
+            <td class="au-u-visible-medium-up">{{format-date generatedForm.created}}</td>
+            <td class="au-u-visible-medium-up">{{format-date-time generatedForm.modified}}</td>
             <td class="au-u-visible-medium-up">
-              <AuLink @route="formbuilder.edit" @model={{row}} @icon="pencil" @iconAlignment="left" @ariaHidden={{true}}>Bewerk</AuLink>
```
Windvis, line 40: We were passing in the "row" object (generatedForm) directly which meant that Ember skipped the model hook of the route. By passing the id we ensure it runs.

app/controllers/formbuilder/edit.js
```diff
@@ -18,10 +18,9 @@ export const GRAPHS = {
 const SOURCE_NODE = new RDFNode('http://frontend.poc.form.builder/sourcenode');
 
 export default class FormbuilderEditController extends Controller {
```
Windvis, line 20: Some methods / tasks / actions can probably be grouped better or reordered. I skipped the cleanup part of the refactor because I ran out of time. 😅

app/components/code-edit-modal.js
JonasVanHoof: Yes indeed I don't see any other tasks that could fix this.. Maybe the `keepLatest` task but it's just going to make everything slower BUT insure that every change in the store is added piece by piece (ignore after the but if its not clear :)) https://ember-concurrency.com/docs/task-concurrency

app/controllers/formbuilder/edit.js
```diff
@@ -108,36 +84,77 @@ export default class FormbuilderEditController extends Controller {
...
   @action
-  serializeSourceToTtl() {
+  async handleCodeChange(newCode) {
+    this.formChanged = true;
+    this.formCode = newCode;
+    this.setupForms();
+  }
+
+  handleBuilderFormChange = restartableTask(async () => {
+    await timeout(1); // small timeout to group multiple store observer callbacks together
```
JonasVanHoof, line 125: I am not getting this comment.. If I read it correctly: this is a restartable task and you want to wait 1ms inbetween restart to make sure the observables are ready?

app/controllers/formbuilder/edit.js
JonasVanHoof: I think it's already very clean ! But I don't know what magic you where thinking of :blush:

JonasVanHoof:
As this is an important change maybe @ZijneDwaasheid  and @aatauil can have a look (if they have time) so they are up to date with the code and we have some more people testing if everything works besides you and me :)

JonasVanHoof:
Something i just saw.. there are some errors being thrown by a `notificationbar` ? Could not find where this was coming from :/ 
image

aatauil:
@JonasVanHoof Cant seem to reproduce that error. Normal flow seems to work so guessing its a flaky error. we can always open an issue if it appears again.

PR #43, opened 2023-10-18
Add more validations per field type

JonasVanHoof:
## ID
 [SFB-39](https://binnenland.atlassian.net/browse/SFB-39)

 ## Description

The `form-builder`  now only has two possible validations. Those are the `RequiredConstraint` & `Maxlength` as there are much more validations than those two I am extending the list of validations that are possible for the different `field types` in this PR. 

In the [ticket above](https://binnenland.atlassian.net/browse/SFB-39) you will find a list of all the field types with the validations I have added in this PR.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Add two or three fields to your form with a different `display-type`. Switch to the validations tab an you will see that the options per `display-type` are different. Try out all the validations that you can find in the dropdown.

Try to break it so our users will not stumble on this error.

 ## Links to other PR's

- /


app/components/rdf-form-fields/validation-concept-scheme-selector.js
```diff
@@ -195,4 +207,14 @@ export default class ValidationConceptSchemeSelectorComponent extends InputField
       new Statement(sourceNode, FORM('grouping'), groupingType, graph),
     ];
   }
+
+  createStatementThatConnectsPathsBetweenFieldAndValidation(
```
JonasVanHoof, line 211: This name needs some improvement..
* createStatementToConnectPaths()
* getStatementToAddFieldPathToValidation

Any idea's?

public/forms/builder/meta.ttl
```diff
@@ -1,125 +1,145 @@
...
+  ext:usesConceptScheme "false"^^xsd:boolean ;
+  ext:canHaveValidation 
+      form:RequiredConstraint,
+      form:MaxLength,
+      form:ValidEmail,
+      form:ValidInteger,
+      form:UriConstraint,
+      form:ValidIBAN ;
+  core:uuid "68ccebe5-84fb-4976-a9b1-28b1d3ef3687" ;
+  skos:inScheme <http://lblod.data.gift/concept-schemes/c5a91bd7-3eb5-4d69-a51b-9bac6bf345f6> ;
```
JonasVanHoof, line 23: I cannot change this to a `prefix` because than no options are shown in the dropdown. This is still the same `uri` as in the selector field in the `form.ttl` of the builder. Not sure why  I can't do this

app/utils/validation-shape-validators.js
```diff
@@ -1,15 +1,33 @@
...
+    store,
+    graph
+  );
+  if (!defaultCountryCodesAdded.isAdded) {
+    const defaultCountryCodeStatements =
+      defaultCountryCodesAdded.subjectsToAddDefaultTo.map((subject) => {
+        return createDefaultCountryCodeStatement(subject, graph);
+      });
+    store.addAll(defaultCountryCodeStatements);
+  }
```
JonasVanHoof, line 21: Want to move it to somewhere else but not sure yet where..

PR #46, opened 2023-10-25
Fix: task throws a cancellation error when an update is done in the field forms

JonasVanHoof:
## ID
 [SFB-44](https://binnenland.atlassian.net/browse/SFB-44)

 ## Description

As we changed that every change in a fields validation form triggers a merging of the field form with the builderForm so the validation will be visible in the preview screen. This throwed an error because the current task for merging was cancelled. (See screenshot in the issue).

As the task is restartable previous task can be cancelled. The issue here was that we are `waiting` in the update of the field form what makes the restartable task obsolete. Removed the async await for this and this fixed the issue.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Create a form with one field and add multiple validations to it
2. Create a form with a second field and see if the validations added to the second field are also applied 

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/add-validations-to-form.js
+++ b/app/components/add-validations-to-form.js
@@ -74,38 +74,6 @@ export default class AddValidationsToFormComponent extends Component {
     this.args.onUpdateValidations(this.savedBuilderTtlCode);
   }
 
-<<<<<<< HEAD
-  getFieldDataForStoreWithForm(storeWithForm) {
-    const isValidTtl = areValidationsInGraphValidated(
-      storeWithForm.store,
-      this.graphs.sourceGraph
-    );
-    if (isValidTtl) {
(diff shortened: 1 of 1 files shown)


JonasVanHoof:
Removed the reviewers as I saw that the result message is not always correct anymore..
image

As you can see the result message is not shown. If you look into the form ttl code the `sh:resultMessage` is also not set on the object. This means that when a change is applied top the form one or maybe more are not applied anymore..

JonasVanHoof:
As the result message is sometimes not added I checked if the flow of the code is still correct and this looks like it is doing ok. 
image

Next possibility is that it is removing items when it is not needed.. let's find out :crossed_fingers:

PR #48, opened 2023-10-26
Refactor: merging of the field forms

JonasVanHoof:
## ID
 [SFB-44](https://binnenland.atlassian.net/browse/SFB-44)

 ## Description

As we changed that every change in a fields validation form triggers a merging of the field form with the builderForm so the validation will be visible in the preview screen. This throwed an error because the current task for merging was cancelled. (See screenshot in the issue).

As the task is restartable previous task can be cancelled. The issue here was that we are `waiting` in the update of the field form what makes the restartable task obsolete.

Reverted to the commit where the valdiations where only applied when the switch was toggled. Wokring from there on to improve the way it works.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Create a form with one field and add multiple validations to it
2. Create a form with a second field and see if the validations added to the second field are also applied 

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/add-validations-to-form.js
+++ b/app/components/add-validations-to-form.js
@@ -60,12 +60,6 @@ export default class AddValidationsToFormComponent extends Component {
     this.registerToObservableForStoresWithForm(this.storesWithForm);
   }
 
-  willDestroy() {
-    super.willDestroy(...arguments);
-
-    this.args.onUpdateValidations(this.savedBuilderTtlCode);
-  }
-
   async mergeFieldDataWithBuilderForm(fieldData) {
     this.isMerging = true;
     const storeWithMergedField = await mergeFieldValidationFormWithBuilderForm(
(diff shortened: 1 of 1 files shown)


JonasVanHoof:
@ZijneDwaasheid  could you check this PR on user flow? The code can also be looked at to see if it's clear what im trying to achieve.

I want to merge this PR asap as this is a revert of the latest commit on master and I don't want you or others to have a merge conflict because of me yeey :)

app/components/add-validations-to-form.js
```diff
@@ -59,62 +60,59 @@ export default class AddValidationsToFormComponent extends Component {
...
+    super.willDestroy(...arguments);
 
-    for (const fieldData of this.getFieldsData(this.storesWithForm)) {
-      const storeWithMergedField = await mergeFieldDataWithBuilderForm(
-        fieldData,
-        this.savedBuilderTtlCode,
-        this.graphs
-      );
+    this.args.onUpdateValidations(this.savedBuilderTtlCode);
+  }
```
JonasVanHoof, line 67: This looks like it can be removed

JonasVanHoof: Removed it

PR #49, opened 2023-10-27
SFB-46: make the code editor workable

JonasVanHoof:
## ID
 [SFB-46](https://binnenland.atlassian.net/browse/SFB-46)

 ## Description

Editing the code in the code editor is atm not workable. In this PR you can find that two buttons are added to reset the code to the previous state or save the current code. 

The big change here is that the code is only updated to the preview and the builder if you press the save button. 

Not a fan of:
* ~when pressing save the editor closes (this is intentionally) => cannot update the this.args.code of the component or should rerender it fully~
* ~When you go directly to the editor without changing anything the code you will see is the `basic-form-template` with the comments. But when you change something in that form, save it and than open the editor again it still shows the code that represents the output but now it is optimized by the generaters. I think this is very strange to the people using this and seeing that the code they just updated is something else.~

TODO:
* Would like to add that you cannot safe the code only reset it when the editor is returning an error.
  - the code mirror editor is not storing or passing errors to the user to display https://discuss.codemirror.net/t/how-should-i-get-the-error-message/6327, https://discuss.codemirror.net/t/showing-syntax-errors/3111/6
* ~new `basic-form-template` code? https://github.com/lblod/frontend-form-builder/pull/49#issuecomment-1782549922~
  -  ~Updated the default template here https://github.com/lblod/frontend-form-builder/pull/50~

 ## Type of change

 - [x] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Edit the code and add a change
2. Reset the code
3. Update the naming of a predicate value
4. Save the code

 ## Links to other PR's

 - [SFB-43](https://binnenland.atlassian.net/browse/SFB-43)

## Attachments:
image

### DIFF
--- a/app/components/code-edit-modal.hbs
+++ b/app/components/code-edit-modal.hbs
@@ -1,13 +1,23 @@
-<AuModal
-  @modalOpen={{true}}
-  @closeModal={{@onClose}}
->
+<AuModal @modalOpen={{true}} @closeModal={{@onClose}}>
   <:title>Formulier code bewerken</:title>
   <:body>
     <div
-      {{editor @code (perform this.handleCodeChange)}}
+      {{editor this.formCode (perform this.handleCodeChange)}}
       {{!template-lint-disable no-inline-styles}}
       style="overflow: auto; height: 100%; width:100%"
     ></div>
+    <div class="au-u-text-right">
+      <AuButton
+        class="au-u-margin-left-small"
+        @skin="secondary"
+        @disabled={{this.isButtonsDisabled}}
+        {{on "click" this.restoreForm}}
+      >Reset</AuButton>
+      <AuButton
+        class="au-u-margin-left-small"
+        @disabled={{this.isButtonsDisabled}}
+        {{on "click" this.updateForm}}
+      >Bewaar</AuButton>
+    </div>
   </:body>
 </AuModal>
--- a/app/components/code-edit-modal.js
+++ b/app/components/code-edit-modal.js
@@ -37,7 +37,6 @@ export default class CodeEditModal extends Component {
   @action
   updateForm() {
     this.formCode = this.formCodeManager.getTtlOfLatestVersion();
-    this.formCodeManager.pinLatestVersionAsReference();
 
     this.args.onCodeChange?.(this.formCode);
     this.updateButtonDisabledState();
--- a/app/components/code-edit-modal.js
+++ b/app/components/code-edit-modal.js
@@ -29,9 +29,7 @@ export default class CodeEditModal extends Component {
 
   @action
   restoreForm() {
-    this.formCode = this.formCodeManager.getTtlOfVersion(
-      this.formCodeManager.getReferenceVersion()
-    );
+    this.formCode = this.formCodeManager.getTtlOfReferenceVersion();
     this.formCodeManager.addFormCode(this.formCode);
     this.updateButtonDisabledState();
   }
@@ -39,7 +37,7 @@ export default class CodeEditModal extends Component {
   @action
   updateForm() {
     this.formCode = this.formCodeManager.getTtlOfLatestVersion();
-    this.formCodeManager.pinLatestVersionAsReferenceTtl();
+    this.formCodeManager.pinLatestVersionAsReference();
 
     this.args.onCodeChange?.(this.formCode);
     this.updateButtonDisabledState();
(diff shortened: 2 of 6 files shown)


JonasVanHoof:
Maybe we should change the ttl code in the `basic-form-template` to the code when it is generated once. This would would fix the `im not a fan of` that the code looks totally different when you first edit it, save it and than look back at the code in the editor.

JonasVanHoof:
Updated the code so the changes in the editor are also managed by the `formCodeManager` service. Found some quirks that when changing something in the builder and than going to the editor the reset button and the save button was not correct. So now the state of the buttons is also managed by checking the code changes through the `formCodeManager` 
 service.

@ZijneDwaasheid  could you check when updating the form (in every usecase) that this is correct?

- builder change => is editor state correct
- editor change => is builder change correct
- builder change => editor change => is editor + builder correct

I think these are the usecases but maybe I forgot one so please check

ZijneDwaasheid (changes requested):
When editing in the code-editor the state of the 'save-button'  isn't  correct. The changes aren't saved the store. So when refreshing changes are gone. 

Maybe change "bewaar" to "aanpassingen toepassen".

Buttons in the editor are only visable when scrolling down (problem for bigger forms).
Maybe pin the buttons so the code scrolls behind it.

app/components/code-edit-modal.hbs
```diff
@@ -1,13 +1,23 @@
...
+        @skin="secondary"
+        @disabled={{this.isButtonsDisabled}}
+        {{on "click" this.restoreForm}}
+      >Reset</AuButton>
+      <AuButton
+        class="au-u-margin-left-small"
+        @disabled={{this.isButtonsDisabled}}
+        {{on "click" this.updateForm}}
+      >Bewaar</AuButton>
+    </div>
```
ZijneDwaasheid, line 21: Maybe pin these buttons

app/components/code-edit-modal.js
```diff
@@ -1,9 +1,49 @@
...
+    this.updateButtonDisabledState();
+  }
+
+  @action
+  updateForm() {
+    this.formCode = this.formCodeManager.getTtlOfLatestVersion();
+
+    this.args.onCodeChange?.(this.formCode);
+    this.updateButtonDisabledState();
+  }
```
ZijneDwaasheid, line 43: Do we want this to only update or actually save the updated code?

ZijneDwaasheid (changes requested):

app/components/code-edit-modal.js
JonasVanHoof: This should only update the form code and not save it to the database. This is done because when the form code includes an error the original form code is not overwritten. And users can just refresh the page to fix the issue

app/components/code-edit-modal.hbs
JonasVanHoof: I could fix the styling right away but as we have a meeting in the afternoon on how to make the application more dev focused I will leave this aside for now because the position can change depending on what comes out of the meeting

JonasVanHoof: Also mentioned it here on [jira](https://binnenland.atlassian.net/browse/SFB-46?focusedCommentId=91282)

JonasVanHoof:
Just found bug that when going to the validations tab and then changing a `validation node` in the code editor that this is not saved as the new builder code is not passed to the validations tab dynamically.

Note: every time something is adjusted in the `code editor` this ttl code is passed to the `builder`. And the validations tab is shown after extracting the fields out of the `builder ttl`. This means that indeed when updating validations in the `code editor` this will not be applied

app/services/form-code-manager.js
```diff
@@ -4,48 +4,59 @@ export default class FormCodeManagerService extends Service {
...
       this.formCodeHistory[this.referenceVersion]
     );
   }
+
+  clearHistory() {
+    this.formCodeHistory = [];
+    this.latestVersion = this.startVersion;
+  }
+
+  #getTtlOfVersion(version) {
```
aatauil, line 52: Typo?
```suggestion
  getTtlOfVersion(version) {
```

aatauil (approved):
It seems to work as expected. We do want to get rid of the service in the next iteration, for now it can stay. Thanks for the fixes! :)

app/services/form-code-manager.js
JonasVanHoof: The method is not used outside of the class so i made it private :) No usecases for it to call it directly atm

PR #50, opened 2023-10-27
feat: form is changed needs an upgrade

JonasVanHoof:
## ID
 SFB-46](https://binnenland.atlassian.net/browse/SFB-46)

 ## Description

When switching between tabs or updating the code the formChanged variable is not updated correctly.  

In this PR im creating a service that will keep track of the version of the ttl code. Every time the code is saved the reference is move to the latest ttl code. Every change made from there on will be compared with the reference.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Change the builder form or validation form so the preview screen  is updated. See if the `formChanged` is correct.
2. Go back to the original state of the form manually and see if `formChanged` is back to `false`

 ## Links to other PR's

 -  https://binnenland.atlassian.net/browse/SFB-43

### DIFF
--- a/app/services/form-code-manager.js
+++ b/app/services/form-code-manager.js
@@ -13,9 +13,6 @@ export default class FormCodeManagerService extends Service {
 
   pinLatestVersionAsReferenceTtl() {
     this.referenceVersion = this.getLatestVersion();
-    console.info(
-      `Updated reference version to latest ${this.referenceVersion}`
-    );
   }
 
   getHistory() {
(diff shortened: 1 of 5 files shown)


JonasVanHoof:
Maybe remove or limit the console logs in between because they are a little to much when you are for example debugging for another issue. Should we only keep the update of the reference log?

app/services/form-code-manager.js
```diff
@@ -0,0 +1,62 @@
+import Service from '@ember/service';
+
+export default class FormCodeManagerService extends Service {
+  startVersion = -1;
+
+  formCodeHistory = [];
+  version = this.startVersion; // -1 so start version is 0
```
JonasVanHoof, line 7: This could possibly be an array of objects so we can start on version 1.
```js
{
  version: number,
  ttl: string
}

JonasVanHoof: But for the usecase now I think we can keep it simple :D

PR #52, opened 2023-11-06
SFB-76: Require name when creating new form

ZijneDwaasheid:
## ID
 SFB-76

 ## Description

We want that when adding a new form we have to name it first.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Click "Maak nieuw semantisch formulier" in the formbuilder app. before you can start building a form, you will be asked to give a name and description (by default "New Form" and "Blank Form" will be filled in). You can click on "maak formulier aan" that will make you go to the edit page or on the little cross that will destroy the record an takes you back to the index page.


app/components/input-new-form-information.hbs
```diff
@@ -0,0 +1,32 @@
...
+          value={{this.formComment}}
+          {{on "input" this.handleCommentChange}}
+        />
+      </div>
+    </div>
+  </Modal.Body>
+  <Modal.Footer>
+    <AuButton {{on 'click' this.updateForm}}>Maak formulier aan</AuButton>
+  </Modal.Footer>
+</AuModal>
```
JonasVanHoof, line 32: Could we also use this component when editing the name and description of the form? Its the same modal only the actions are different but that can be solved with a callback maybe

JonasVanHoof (commented):
Could this modal be showed over the formbuilder? The background white is a little abrupt and it's not that the content shown behind should change.
image

FIXED

JonasVanHoof:
Should the popup have a default name? When people do not want to name it they are just going to click over it and that will result in again the issue that they are named the same. quote: `this should be the case where you get a modal that then requires you to name the form`

Also should there be a check that you cannot name a form the same as an existing one?

aatauil:
@JonasVanHoof @ZijneDwaasheid To clarify. The name of a form is both required and needs to be unique.

app/routes/formbuilder/new.js
```diff
@@ -10,15 +10,11 @@ export default class FormbuilderNewRoute extends Route {
     const newForm = await this.store.createRecord('generated-form', {
```
aatauil, line 10: This code runs on route load. This means that it creates a new empty form before the user specified any title or description

app/components/input-new-form-information.hbs
aatauil, line 1: No need to make the modal into seperate component. This will only be used in one place.

app/components/input-new-form-information.hbs
```diff
@@ -0,0 +1,50 @@
...
+    <div class="au-o-grid au-o-grid--small">
+      <div class="au-o-grid__item au-1-1">
+        <AuLabel for="input-filelabel"
+          @required={{true}}
+        >
+          Naam
+        </AuLabel>
+        <AuInput
+          id="input-filelabel"
+          @error={{this.emptyFormLabel}}
```
aatauil, line 15: We only want to show an error after the user has performed an action.

app/components/input-new-form-information.hbs
```diff
@@ -0,0 +1,50 @@
...
+          value={{this.formLabel}}
+          {{on "change" this.handleLabelChange}}
+        />
+        {{#if this.emptyFormLabel}}
+          <AuHelpText @error={{true}}>Een naam geven is verplicht</AuHelpText>
+        {{/if}}
+      </div>
+      <div class="au-o-grid__item au-1-1">
+        <AuLabel for="input-filecomment"
+          @required={{true}}
```
aatauil, line 25: Description should not be required

aatauil:
@ZijneDwaasheid This PR is not quite ready yet. It needs some refactoring and structural changes.

**comments:**
- While we do want to use a modal, it does not really make sense to show that modal in a new route. Keep the modal in the index route witth the table. 
- remove formbuilder-new route
- Only show input error after the user interacted with the input

aatauil (changes requested):

app/controllers/index.js
```diff
@@ -1,7 +1,85 @@
...
 export default class IndexController extends Controller {
   sort = '-created';
   page = 0;
   size = 20;
+
+  @service store;
+  @service router;
+
+  @tracked showModal = false;
+  @tracked hasBeenFocused = false;
```
JonasVanHoof, line 15: This is tracked but not used in the template?

Is it possible to use the value showModel for this instead of a new one?

JonasVanHoof (changes requested):

JonasVanHoof:
@aatauil  the error message not showing on intially showing the model should work or I'm checking it in another way 
image

app/controllers/index.js
Windvis, line 1: Instead of adding the logic to the controller, I would create a new `CreateFormModal` component. That component could house the template and most of the logic, the controller would then only need a flag to show the modal when needed. 

The main benefit is that we no longer need to do any data resetting when the modal closes, which is error prone. I noticed that the current version shows the `Deze naam is al eens gebruikt` message sometimes when opening and closing the modal several times, because the `duplicateNames` array isn't cleared 😄.

I think that would simplify the code a bit.

app/routes/formbuilder/new.js
```diff
@@ -5,13 +5,15 @@ export default class FormbuilderNewRoute extends Route {
   @service router;
   @service store;
 
-  async model() {
+  async model(params, transition) {
```
Windvis, line 8: If we go the custom component way, I think we can just remove this route altogether. The creation and redirecting could also be part of that component.

It feels a bit strange to have validation logic in a modal while it's still possible to sidestep it by going to the route directly. So I think keeping this together would be "better".

app/templates/index.hbs
```diff
@@ -57,3 +60,51 @@
...
+  @closeModal={{this.closeModal}} as |Modal| >
+  <Modal.Body>
+    <div class="au-o-grid au-o-grid--small">
+      <div class="au-o-grid__item au-1-1">
+        <AuLabel for="input-filelabel"
+          @required={{true}}
+        >
+          Naam
+        </AuLabel>
+        <AuInput
```
Windvis, line 76: Would it make sense to make this field fill the full width in case users want to type a longer name?

app/templates/index.hbs
```diff
@@ -57,3 +60,51 @@
...
   </div>
 </main>
+
+<AuModal
+  @title="Geef een naam en beschrijving"
+  @modalOpen={{this.showModal}}
+  @closeModal={{this.closeModal}} as |Modal| >
+  <Modal.Body>
+    <div class="au-o-grid au-o-grid--small">
+      <div class="au-o-grid__item au-1-1">
```
Windvis, line 70: Nitpick, but it seems the grid is only used to provide some spacing between the input and the textarea. I think it would be better to either use the [`AuFormRow` setup](https://appuniversum.github.io/ember-appuniversum/?path=/story/patterns-forms--default-form) or the [flow utils](https://appuniversum.github.io/ember-appuniversum/?path=/story/objects-flow--page) instead.

app/templates/index.hbs
```diff
@@ -57,3 +60,51 @@
...
+        <AuLabel for="input-filelabel"
```
Windvis, line 71: Nitpick, I would use more descriptive ids. These seem copy pasted from somewhere unrelated to forms 😄.
```suggestion
        <AuLabel for="form-name"
```

app/controllers/index.js
```diff
@@ -1,7 +1,74 @@
 import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
```
Windvis, line 2: FYI, in Ember 4.1+ a new import is available which is shorter:

```suggestion
import { service } from '@ember/service';
```

app/controllers/index.js
```diff
@@ -1,7 +1,74 @@
...
+
+  @service store;
+  @service router;
+
+  @tracked showModal = false;
+  @tracked name = '';
+  @tracked description = ``;
+  @tracked duplicateNames = [];
+
+  modelName = 'generated-form';
```
Windvis, line 19: I don't think this is needed / used?

app/controllers/index.js
```diff
@@ -1,7 +1,74 @@
...
+  closeModal() {
+    this.name = '';
+    this.description = ``;
+    this.showModal = false;
+    this.hasBeenFocused = false;
+  }
+
+  @action
+  async handleNameChange(event) {
+    this.name = event.target.value;
```
Windvis, line 32: I would trim the value before checking for duplicates. Otherwise they might sidestep the validation by adding spaces.

Windvis (commented):
Left some suggestions, but it mostly works as expected 👍.

app/controllers/index.js
ZijneDwaasheid: @aatauil just to check so we don't go back and forward on the custom createFormModal component. If I'm correct you weren't a fan of using a custom component for this modal, because it was only used in this one place.

Windvis: 😄 Fair enough. I don't think keeping it in a controller because it would only be used once is a good argument. I see it more as extracting logic into a util function to make the code more readable. 

Anyway, if that was the previous feedback, keep it like this for now. Make sure all state is properly cleaned up though 😄. I would also still remove the `new` route and do the logic in the controller as well.

app/routes/formbuilder/new.js
Windvis: As mentioned in the other comment, moving this to the index controller keeps things together and makes this route unnecessary.

Windvis (commented):
You didn't have to do the component change; but since you did, one last remark, sorry 😄.

app/components/create-form-modal.js
```diff
@@ -0,0 +1,85 @@
...
+  @tracked duplicateNames = [];
+  @tracked hasBeenFocused = false;
+
+  @action
+  closeModal() {
+    this.args.closeModal();
+    this.name = '';
+    this.description = '';
+    this.duplicateNames = [];
+    this.hasBeenFocused = false;
```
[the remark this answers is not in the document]
ZijneDwaasheid: will do thanks

PR #53, opened 2023-11-07
SFB-75: Refactor current app to work with routes

JonasVanHoof:
## ID
 [SFB-75](https://binnenland.atlassian.net/browse/SFB-75)

 ## Description

 We want to move to show each action on the form into a separate `route` so we do not need to fake reload the components so the data in there is updated. This is a ticket that has to be done before we add the editor to the tabs [SFB-64](https://binnenland.atlassian.net/browse/SFB-64)

**Update:** the logic as changed a bit as we where always having problems with merging the `field validation form` back to the builder without data loss. As it is now you can select the field that you want a validation for and add it. This will also make the list of fields alot shorter and reduce scrolling in the app.

 ## Type of change

 - [X] Refactor
 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Everything should work as in the previous version.

 ## Links to other PR's

- /


JonasVanHoof:
Would like to change this .. to be more somehting like in this [image](https://dribbble.com/shots/18947078-Form-Pills) so you can easily select your (section and) field
image

ZijneDwaasheid:
> Would like to change this .. to be more somehting like in this [image](https://dribbble.com/shots/18947078-Form-Pills) so you can easily select your (section and) field image
> 
> Edit: Don't think this is to hard as the [`AUPill`](https://appuniversum.github.io/ember-appuniversum/?path=/story/components-notifications-aupill--component) is available and than we only have to set the logic showing the data. But want to do this in a separate PR as actually this change should not have been here in the first place... did the change because it makes it alot more simple to understand whats going on.
> 
> Else we should add a section dropdown next to the field dropdown so the user can see the fields per section

The pill thingy could be nice, although for large forms it could get a little chaotic (like 50 fields, idk if forms like that exists).

app/components/add-validations-to-form.hbs
```diff
@@ -1,14 +1,30 @@
...
+        <AuButton
+          {{on "click" (fn this.setSelectedField field)}}
+          @skin="link"
+          role="menuitem"
+        >
+          {{field.name}}
+        </AuButton>
+      {{/each}}
+
+    </AuDropdown>
```
ZijneDwaasheid, line 15: I tested this on the Big Form (50 fields), not everything fits on the dropdown and dropdown isn't scroll-able

ZijneDwaasheid (commented):
Apart of the dropdown not scroll-able and that resulting in a part of the dropdown being off screen (only visable if you zoom out in the browser :) ), functionality wise everything seems to be working. Really like the new tabs with their little icons :)

For me the dropdown menu to edit validations makes it way more easy to access the right field and edit it.

app/components/add-validations-to-form.hbs
JonasVanHoof: I took this form as a standard form in my head https://dev.form-builder.lblod.info/formbuilder/652661D898CA4C0008000019/edit FYI

JonasVanHoof:
@aatauil  the selection of the field in the form should be a different PR. If we keep current implementation we have to create a new jira ticket explaining this feature

aatauil:
@JonasVanHoof Very nice change. We might still want to refactor the previewForm to be in the parent route to limit duplication. Also the tabs dissapear when the codeEditor view becomes larger then the screenheight. It shouldnt block this PR though, will create seperate tickets for these issues. Very nice work

PR #54, opened 2023-11-10
Sfb 64/improve visibility of code editor

JonasVanHoof:
## ID
 [SFB-64](https://binnenland.atlassian.net/browse/SFB-64)

 ## Description

When you want to edit the ttl code directly you had to go to the toolbar and select edit code. As discussed we want this to be in a tab next to the builder and the validations so it is more accessible for the developers.

 ## Type of change

 - [x] Improvement
 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Look if the option is in the toolbar

 ## Links to other PR's

 - http://github.com/PR/2314135

### DIFF
--- a/app/components/toolbar.hbs
+++ b/app/components/toolbar.hbs
@@ -11,7 +11,8 @@
         <ul class="au-c-list-horizontal au-u-padding-right-tiny">
           <li class="au-c-list-horizontal__item">
             <span class="au-c-app-chrome__status">
-              Bewaard op {{or @model.modified @model.created}}
+              Bewaard op
+              {{or @model.modified @model.created}}
             </span>
           </li>
         </ul>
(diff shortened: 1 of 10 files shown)


JonasVanHoof:
The only thing im missing here with the new side panel that we should add tooltips when the user hovers over it. Maybe with three tabs that this is not needed yet but if we are expanding from three to more this would be handy

JonasVanHoof:
Im aware that the scrolling is not working atm..

JonasVanHoof:
Just something I thought this morning and can be easely updated with two AU classes  is to move the navigation to the center so all the items are a bit more spread instead of everythin in the left corner
image

Add the two flex classes to it 
``` 
 <AuTabs
    class="au-u-padding-left-small au-u-padding-right-small au-u-flex au-u-flex--center"
    @reversed={{this.reversed}}
    as |Tab|
  >
  ```

aatauil (approved):

PR #55, opened 2023-11-13
SFB-71: Fix codelijsten

JonasVanHoof:
## ID
[ SFB-71](https://binnenland.atlassian.net/browse/SFB-71)

 ## Description

When selecting the input type `dropdown` it didn't show the options for that `concept-scheme`. This was because we do not have all the options in our metaGraph in the local store. Thisdata has to be get in our Virtuoso store. I created a component that will query the database.

As the styling of the `codelijsten` page was not great (I alays copied `code:`)  with it what results in wrong config for the component. Updated this with use of `AuPanel`.

The options will show in the preview but not sure what has to be done when the user has selected that an option. This component is also only in the `form-builder` so if people want to use this component outside of form builder they will NOT be able to for now.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Go to `codelijsten` endpoint select an item and see if the options display on the right. The code configuration for that `concept-scheme` should show in the panel below. Copy the code and past it in the `configuration textArea`  and than select field type to be `dropdown`.

 ## Links to other PR's

 -


aatauil:
Error in console:
image

app/routes/codelijsten.js
```diff
@@ -1,3 +1,9 @@
 import Route from '@ember/routing/route';
 
-export default class CodelijstenRoute extends Route {}
+export default class CodelijstenRoute extends Route {
+  setupController(controller, model) {
+    super.setupController(...arguments);
+
+    controller.setup(model);
```
aatauil, line 7: avoid using the setupController to act as an init lifecyclehook, better place to do this is the constructor of the controller itself. also specifcally in this case, the setup function does not expect any arguments

JonasVanHoof:
> Error in console: image

Im aware of this.. there we have this issue for [SFB-45](https://binnenland.atlassian.net/browse/SFB-45) on Jira but as the functionality is now it works. The bigger problem is that we have to handle the errors coming from the RDF form/helper package/ store when a change is applied and it is invalid. (Atleast I think it is why the error is popping-up).

Will investigate :ok_hand:

app/routes/codelijsten.js
JonasVanHoof: The constructor is only called initially what means when leaving the `codelijsten` route and than coming back it is not called unfortunately. Also the [glimmer component](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_lifecycle-and-properties) only has two lifecycle hooks to use.. Fixed this by calling the `@action updateSelectedSchemeTo` in the template so it is called everytime the template loads.

JonasVanHoof:
@aatauil the error is not visible anymore and im checking on json format in the component now. Not showing a toaster message is actually not a possibility as this would be more handy for the user than a `console.log` message.

The problem possibly has something to do with the ember array that is being used in the `ToasterService`  [here](https://github.com/appuniversum/ember-appuniversum/blob/master/addon/services/toaster.js#L7C27-L7C27). I updated that line in the ember-appuniversum project to just `@tracked toasts = A();` instead of `@tracked toasts = A([]);` this didn't break anything.

Tried using the appuniversum local package in the form-builder application but this still returns issues with compatibility with version V1 and V2. So after trying i that and not getting it to work I changed it here to just a console.log message.

This is something that has to change but im not sure where the problem actually is...

app/components/store-concept-scheme-selector.js
```diff
@@ -0,0 +1,65 @@
...
+      );
+    } else {
+      console.warn(`Key "conceptScheme" not found in json configuration`);
+    }
+  }
+
+  @action
+  setSelected(value) {
+    this.selected = value;
+    // What should happen here?
```
aatauil, line 40: Nothing, since this will be shown in the previewform :)

app/routes/formbuilder/edit.js
```diff
@@ -70,6 +71,11 @@ export default class FormbuilderEditRoute extends Route {
           'http://lblod.data.gift/display-types/countryCodeConceptSchemeSelector',
         edit: CountryCodeConceptSchemeSelectorComponent,
       },
+      {
+        displayType:
+          'http://lblod.data.gift/display-types/storeConceptSchemeSelector',
+        edit: StoreConceptSchemeSelectorComponent,
```
aatauil, line 77: The use of a custom component might be a bit troublesome. When the user chooses a dropdown, it will reflect in the outputed code. 

image

I tried to overwrite the component but unfortunatly ember-submission-form-fields does not allow it so we might be a bit stuck here.

aatauil (changes requested):
While this is a good solution, it will most likely come back to haunt us later unless we fix the root cause. This will probably need a change in ember-submission-form-fields instead of in the formbuilder

JonasVanHoof:
> While this is a good solution, it will most likely come back to haunt us later unless we fix the root cause. This will probably need a change in ember-submission-form-fields instead of in the formbuilder

@aatauil 
I created this [PR in ESFF](https://github.com/lblod/ember-submission-form-fields/pull/156) to validate the options passed to the `concept-scheme-selector` .

This is only step one of the solution.. are we able to call the database store from the ESFF? Else we could add a boolean `ext:fetchLocal false` to our selector turtle config what will set the way of fetching information in the component. Not sure if this is possible.

app/routes/formbuilder/edit.js
JonasVanHoof: If we..

The code in the editor will always show the custom component `StoreConceptSchemeSelector`. If we add somewhere near the editor a button to export the code we than can overwrite every custom component if we want but not sure if this is something we want to do

app/utils/query-sparql-query.js
```diff
@@ -0,0 +1,16 @@
+export async function queryDB(query, sparqlEndpoint = '/sparql') {
```
JonasVanHoof, line 1: I created this method as it should return the json response of the query. I also saw that [here](https://github.com/lblod/frontend-form-builder/blob/master/app/services/database.js) that there is a database service that does almost the same with the difference that that method returns the full response and not the bindings as json. I could add ... this after my query but that would make it very ugly and not easy to understand.
```
let jsonResponds = await response.json();
this.options =  jsonResponds.results.bindings;
```

Just adding this comment here as we should be aware of this and maybe we can find a solution :)

app/components/store-concept-scheme-selector.js
```diff
@@ -0,0 +1,50 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { queryDB } from '../utils/query-sparql-query';
+
+export default class StoreConceptSchemeSelectorComponent extends Component {
```
JonasVanHoof, line 6: WE should be creating the custom components for the radio & checkbox selectors to if we implement the overwrite method for the preview.

JonasVanHoof:
Will be removing the overwrite method and the custom component to show the options in the selectors. First will be implementing the backend and updating the frontend to work with the store instead of direct sparql queries. More info can be found here https://github.com/lblod/frontend-form-builder/pull/66

JonasVanHoof:
Closing this PR because of implemetation for the concept-schemes by store. https://github.com/lblod/frontend-form-builder/pull/66

PR #57, opened 2023-11-14
SFB-81: Implement feature flags

JonasVanHoof:
## ID
 [SFB-81](https://binnenland.atlassian.net/browse/SFB-81)

 ## Description

 This PR  will make it possible to use feature flags for the application. Example in the future we want a `codelijsten` in the application but we do not want to show it yet on our online application.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Go to `config/environment.js` and change the boolean under `featureFlags` to false. now you will see that the builder/validation switch will appear or dissappear.

 ## Links to other PR's

 - /

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -18,6 +18,7 @@ module.exports = function (environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
+    featureFlags: {},
   };
 
   if (environment === 'development') {
@@ -26,6 +27,7 @@ module.exports = function (environment) {
     // ENV.APP.LOG_TRANSITIONS = true;
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
+    ENV.LOG_FEATURE_FLAG_MISS = true;
   }
 
   if (environment === 'test') {
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,6 +42,7 @@
         "ember-concurrency": "^3.1.1",
         "ember-data": "~4.12.0",
         "ember-data-table": "^2.0.0",
+        "ember-feature-flags": "^6.0.0",
         "ember-fetch": "^8.1.2",
         "ember-load-initializers": "^2.1.2",
         "ember-modifier": "^3.2.7",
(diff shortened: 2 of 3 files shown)


JonasVanHoof:
How should you add a feature flag?

1. If the feature you want to hide behind a flag is under a `route` that does not have the service injected
    * inject the service `features` into the route
    * return the service in the model
2. In your `.hbs` file you can use `this.model.features.<yourAmazingfeatureName>` 
3. In your `.js` file you can access it by writing `this.model.features.isEnabled('<yourAmazingfeatureName>')`
4. Make sure the applicaiton is refreshed and you will see the result depending on the config

app/controllers/formbuilder/edit.js
```diff
@@ -89,6 +89,9 @@ export default class FormbuilderEditController extends Controller {
   });
 
   setupPreviewForm = restartableTask(async () => {
+    if (this.model.features.isEnabled('showLogWhenPreviewIsBeingrefreshed')) {
+      console.log(`Refreshing preview`);
+    }
```
JonasVanHoof, line 94: Remove*

app/templates/formbuilder/edit.hbs
```diff
@@ -11,14 +11,19 @@
...
+        <AuToggleSwitch
+          @disabled={{or
+            this.setupBuilderForm.isRunning
+            this.setupPreviewForm.isRunning
+          }}
+          @checked={{not this.isShowBuilder}}
+          @onChange={{this.toggleIsAddingValidationToForm}}
+        >Validation</AuToggleSwitch>
+      </div>
+    {{/if}}
```
JonasVanHoof, line 26: Remove the if around it

config/environment.js
```diff
@@ -18,6 +18,10 @@ module.exports = function (environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
+    featureFlags: {
+      showBuilderValidationSwitch: false,
+      showLogWhenPreviewIsBeingrefreshed: false,
```
JonasVanHoof, line 23: *remove the flags

app/routes/formbuilder/edit.js
```diff
@@ -25,6 +26,7 @@ export default class FormbuilderEditRoute extends Route {
       generatedForm,
       formTtl,
       metaTtl,
+      features: this.features,
```
wolfderechter, line 29:
```suggestion

```

app/templates/formbuilder/edit.hbs
```diff
@@ -11,14 +11,19 @@
...
   <div class="au-c-main-container__content">
-    <div class="playground-tab">
-      Builder
-      <AuToggleSwitch
-        @disabled={{or this.setupBuilderForm.isRunning this.setupPreviewForm.isRunning}}
-        @checked={{not this.isShowBuilder}}
-        @onChange={{this.toggleIsAddingValidationToForm}}
-      >Validation</AuToggleSwitch>
-    </div>
+    {{#if this.model.features.showBuilderValidationSwitch}}
```
wolfderechter, line 14:
```suggestion
    {{#if this.features.showBuilderValidationSwitch}}
```

wolfderechter (commented):
I think things can be a bit simpler by just using `this.features...` instead of the `this.model.features`

JonasVanHoof:
> I think things can be a bit simpler by just using `this.features...` instead of the `this.model.features`

I agree that we should just inject the service in the component or item we are adding the feature to. This limits the scope.

For this PR: will remove the examples and just leave the adding o the package and in the `environment.js` add `featureFlags: {}`

JonasVanHoof:
> > ```
> > 3. In your `.js` file you can access it by writing `this.model.features.isEnabled('<yourAmazingfeatureName>')`
> > ```
> 
> We can also just use `this.features.get('<yourAmazingfeatureName>')` for js

I you read the code `this.features.get('<yourAmazingfeatureName>')` it is less descriptive IMO than `this.features.isEnabled('<yourAmazingfeatureName>')` but that's preference I think. In the future when we start adding implementations behind feature flags we will flashback on this on how to work with it best.

PR #59, opened 2023-11-16
SFB-82: Duplicate preview form in each route template

JonasVanHoof:
## ID
 [SFB-82](https://binnenland.atlassian.net/browse/SFB-82)

 ## Description

In the navigation at the top are three routes defined. Each route has ts own preview rdf-form and this should'nt be the case. We will be moving this rdf-form to the parent `edit` template so it is common and we do not have duplicate code that will render the preview form.

 ## Type of change

 - [x] Refactor
 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Everything should work as before. In the code there should only be one `rdf-form` with a `previewStore`

 ## Links to other PR's

 - /


aatauil (approved):
No duplication no more. Very nice, thanks :+1:

PR #60, opened 2023-11-20
SFB-85: validations form data not always shown

JonasVanHoof:
## ID
 [SFB-85](https://binnenland.atlassian.net/browse/SFB-85)

 ## Description

When selecting a field in the validations tab the data is not always loaded. Now the data is created when selecting a field which means the data should always bet there when asked for.
 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Open the validation tab and start switching between fields and while doing so add and remove some validations.

 ## Links to other PR's

 - /


ZijneDwaasheid (approved):
OK this fixes the bug, validations are loaded when switching between selected fields.

PR #61, opened 2023-11-23
SFB-88: Validations cannot be edited of V2 form 

JonasVanHoof:
## ID
[SFB-88](https://binnenland.atlassian.net/browse/SFB-88)

 ## Description

 The `form-builder` adds validations as separate nodes to the `ttl code` and manipulates that code. But V2 forms are using `blank node`  what the `form-builder` does not understand.  The overal use of validations is in `blank nodes` so in this PR I ''ll be updating the code that the validations will be added and  updated not as `named nodes` but as `blank nodes`

 ## Type of change

 - [x] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Add and remove validations on a field. Switch fields and see if the validations are only applied to the specific field. Check the outcome `ttl code` in the editor. Than update the code in the `editor` and see if the change is applied to the `builder` and `validations` tab

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/add-validations-to-form.js
+++ b/app/components/add-validations-to-form.js
@@ -48,8 +48,13 @@ export default class AddValidationsToFormComponent extends Component {
     this.builderStore = new ForkingStore();
     yield parseStoreGraphs(this.builderStore, ttlCode);
 
-    this.fields = yield this.createSeparateStorePerField(this.builderStore);
-    this.fields.length >= 1 ? (this.selectedField = this.fields[0]) : null;
+    this.fields = this.createFieldArray(this.builderStore);
+    this.fields.length >= 1
+      ? this.setSelectedField({
+          name: this.fields[0].name,
+          subject: this.fields[0].subject,
+        })
+      : null;
   }
 
   @action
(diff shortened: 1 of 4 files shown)


JonasVanHoof:
@ZijneDwaasheid 

Finally fixed the duplicate issue! See last commit that was the problem.. raather logical tbh.. BUT try then code add and remove validations when started from a new form + create a form with this ttl code 
[Wolfs_fixed_copy_paste_form.txt](https://github.com/lblod/frontend-form-builder/files/13449464/Wolfs_fixed_copy_paste_form.txt) a form that He wanted to show and edit and it was not possible.

ZijneDwaasheid (commented):
OK two things I encountered: 
1) when there is only 1 field, you can still click the dropdown and get an error because you have to have at least 2 fields (the one you are currently edting the validation of and one in the dropdown). If there is only 1 field, the dropdown is empty and expects at least one field.

2) Here I am again with Urban Renewal, the validation tab doesn't work because it thinks there are no fields, the builder tab and preview think otherwise, that a 2-1 :)
I tried this: https://dev.form-builder.lblod.info/formbuilder/655DBE0E98CA4C0008000028/edit/code (just copy-paste)

Or maybe this form is cursed

JonasVanHoof:
> OK two things I encountered:
> 
> 1. when there is only 1 field, you can still click the dropdown and get an error because you have to have at least 2 fields (the one you are currently edting the validation of and one in the dropdown). If there is only 1 field, the dropdown is empty and expects at least one field.
> 2. Here I am again with Urban Renewal, the validation tab doesn't work because it thinks there are no fields, the builder tab and preview think otherwise, that a 2-1 :)
>    I tried this: https://dev.form-builder.lblod.info/formbuilder/655DBE0E98CA4C0008000028/edit/code (just copy-paste)
> 
> Or maybe this form is cursed

Alright the first one i didn't notice good catch! Already pushed a fix for that :) The second one I see that this is not finding the fields in the form but actually the issue description tells us that the validations cannot be edited because they are blank nodes 😄  I can definitely see why this bug can also fall under this issue but could we create different PR for this ? Than the logic that now blank nodes are created is implemented and we can look into the fields not being recognized in a different PR

ZijneDwaasheid (approved):

PR #63, opened 2023-11-27
SFB-88: V2 form fields are visible and can be edited in the validations tab 

JonasVanHoof:
## ID
[SFB-88](https://binnenland.atlassian.net/browse/SFB-88)

 ## Description

 The `form-builder` adds validations as separate nodes to the `ttl code` and manipulates that code. But V2 forms are using `blank node`  what the `form-builder` does not understand.  The overal use of validations is in `blank nodes` so in this PR I ''ll be updating the code that the validations will be added and  updated not as `named nodes` but as `blank nodes`

 ## Type of change

 - [x] Bug fix
 - [] New feature
 - [] Breaking change
 - [ ] Other

 ## How to test

Add and remove validations on a field. Switch fields and see if the validations are only applied to the specific field. Check the outcome `ttl code` in the editor. Than update the code in the `editor` and see if the change is applied to the `builder` and `validations` tab

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/add-validations-to-form.hbs
+++ b/app/components/add-validations-to-form.hbs
@@ -1,33 +1,37 @@
 <div>
-  <PowerSelect
-    @searchField="name"
-    @searchEnabled={{true}}
-    @selected={{this.selectedField}}
-    @options={{this.fields}}
-    @onChange={{this.setSelectedField}}
-    @allowClear={{true}}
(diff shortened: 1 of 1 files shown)


ZijneDwaasheid (commented):
works on Urban Renewal V2 form

ZijneDwaasheid (approved):

PR #65, opened 2023-11-28
SFB-65: Will add an empty line after every dot

JonasVanHoof:
## ID
 [SFB-65](https://binnenland.atlassian.net/browse/SFB-65)

 ## Description

The editor code output is not always so amazing. Add a new line after a dot, a new line and a tab after a semi-colon.
 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Write some code an lose focus out of the editor and see if it formats.

 ## Links to other PR's

 - /


ZijneDwaasheid (commented):
Could it be that you have to click in the editor (it does some changes not all) and them out of the editor (now all changes happened) .

JonasVanHoof:
> Could it be that you have to click in the editor (it does some changes not all) and them out of the editor (now all changes happened) .

Running the `empty line after a line that ends with a dot ` twice so the code get's cleaned up. I think this should be enough for that change.

JonasVanHoof:
Found an [app](https://github.com/replit/codemirror-vscode-keymap/blob/master/src/index.ts) that uses alot of vs code shortcuts but I think  the ones we are using are already in the basicSetup available

aatauil:
@JonasVanHoof This PR does too much, would have loved to see it split into 2 PR's instead
- format code
- add macros/completions

For now its ok

JonasVanHoof:
> @JonasVanHoof This PR does too much, would have loved to see it split into 2 PR's instead
> 
> * format code
> * add macros/completions
> 
> For now its ok

Reoccuring note to my PR's, sorry. Added a note to self at the top of my daily notes file ;p 
image

Windvis (commented):
I noticed that the `Alt-i` keybind doesn't work on macOS. It outputs the `ˆ` instead so this is probably an OS thing that can't easily be changed

Is there a specific reason why `Alt-i` was used? Is that the keybind from a certain editor?

I just checked VS Code on and it seems to use `Shift-Alt-f` on Mac. I'm not sure what it is on Windows / Linux, but maybe we can use that here as well? Codemirror also has an option for OS specific keybinds so we can use a different one for mac if we want: https://codemirror.net/docs/ref/#view.KeyBinding.mac

Once I modified the keybind everything seemed to work as expected 👍.

Windvis:
> How to test
> Write some code an lose focus out of the editor and see if it formats. 

Just double checking, but I think it doesn't do any auto-formatting at the moment so these instructions are incorrect? Or is that supposed to work? 😄 I could only get it to format with a keybind.

JonasVanHoof:
> > How to test
> > Write some code an lose focus out of the editor and see if it formats.
> 
> Just double checking, but I think it doesn't do any auto-formatting at the moment so these instructions are incorrect? Or is that supposed to work? 😄 I could only get it to format with a keybind (which is fine, IMO).

Very true, it was implemented with the lose focus in the beginning but we thought that a shortcut fits better for us. I updated the test description. Thanks!

JonasVanHoof:
> I noticed that the `Alt-i` keybind doesn't work on macOS. It outputs the `ˆ` instead so this is probably an OS thing that can't easily be changed
> 
> Is there a specific reason why `Alt-i` was used? Is that the keybind from a certain editor?
> 
> I just checked VS Code and it seems to use `Shift-Alt-f` on Mac. I'm not sure what it is on Windows / Linux, but maybe we can use that here as well? Codemirror also has an option for OS specific keybinds so we can use a different one for mac if we want: https://codemirror.net/docs/ref/#view.KeyBinding.mac
> 
> Once I modified the keybind everything seemed to work as expected 👍.

Updated it to the keys used for max windows and linux https://github.com/lblod/frontend-form-builder/pull/65/commits/a2b88f32a2db9e11018a478abf32188968e146f3

vscode windows:  https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf
vscode mac: https://code.visualstudio.com/shortcuts/keyboard-shortcuts-macos.pdf

I assume that the shortcut u use represents this in the shortcut guide `⇧⌥F F` 😁

PR #66, opened 2023-12-06
SFB-100: create concept-scheme and concept model from store instead of direct sparql

JonasVanHoof:
## ID
 [SFB-100](https://binnenland.atlassian.net/browse/SFB-100)

 ## Description

The concept schemes are fetched with a direct sparql query in the frontend. Change this by getting the concept-schemes with their concepts to be get from the store `this.store.query('concept-scheme')`
 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Checkout the app for the frontend on the the `SFB-100` branch. Make sure all the migrations have run.

Check the code if the `concept-scheme-uri-selector` is using the store instead of a direct sparql query. All options should be visible with their concepts on the left side.

 ## Links to other PR's

 - https://github.com/lblod/app-form-builder/pull/7

### DIFF
--- a/app/components/concept-scheme-uri-selector.js
+++ b/app/components/concept-scheme-uri-selector.js
@@ -16,15 +16,19 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
 
   @action
   async loadOptions() {
-    const conceptSchemes = await this.store.query('concept-scheme', {
-      include: 'concepts',
-    });
+    const conceptSchemes = await this.store.query('concept-scheme', {});
 
     this.options = this.getSortedOptions(conceptSchemes);
   }
 
   async update() {
-    const concepts = await this.selected.concepts;
+    const concepts = await this.store.query('concept', {
+      filter: {
+        'concept-schemes': {
+          ':uri:': this.selected.uri,
+        },
+      },
+    });
 
     this.args.update({
       uri: this.selected.uri,
--- a/app/components/concept-scheme-uri-selector.js
+++ b/app/components/concept-scheme-uri-selector.js
@@ -1,10 +1,11 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-
-const SPARQL_ENDPOINT = '/sparql';
+import { inject as service } from '@ember/service';
 
 export default class ConceptSchemeUriSelectorComponent extends Component {
+  @service store;
+
   @tracked selected;
   @tracked options;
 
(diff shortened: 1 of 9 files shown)


app/components/concept-scheme-uri-selector.js
```diff
@@ -13,26 +14,25 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
...
   async loadOptions() {
-    this.options = await this.queryDB(`
-    SELECT DISTINCT * {
-      ?uri a <http://www.w3.org/2004/02/skos/core#ConceptScheme>;
-        <http://www.w3.org/2004/02/skos/core#prefLabel> ?prefLabel.
-    }
-  `);
+    const conceptSchemes = await this.store.query('concept-scheme', {});
+
+    this.options = this.getSortedOptions(conceptSchemes);
```
JonasVanHoof, line 21: * updated

app/components/concept-scheme-uri-selector.js
```diff
@@ -13,26 +14,25 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
...
+          ':uri:': this.selected.uri,
+        },
+      },
+    });
 
     this.args.update({
-      uri: this.selected.uri.value,
-      concepts: concepts.map((concept) => concept.prefLabel.value),
+      uri: this.selected.uri,
+      concepts: [...concepts].map((concept) => concept.label),
```
JonasVanHoof, line 35: * updated

app/components/concept-scheme-uri-selector.js
```diff
@@ -42,20 +42,15 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
...
-      headers: { Accept: 'application/sparql-results+json' },
+  getSortedOptions(conceptSchememodels) {
+    return [...conceptSchememodels].sort(function (a, b) {
+      if (a.label < b.label) {
+        return -1;
+      }
+      if (a.label > b.label) {
+        return 1;
+      }
+      return 0;
```
JonasVanHoof, line 53: Can this be fixed by adding a sort on label in the store query?

app/models/concept-scheme.js
```diff
@@ -0,0 +1,11 @@
+import Model, { attr, hasMany } from '@ember-data/model';
+
+export default class ConceptSchemeModel extends Model {
+  @attr uri;
+  @attr preflabel;
+  @hasMany('concept', { inverse: null, async: true }) concepts;
+
+  get label() {
+    return this.preflabel;
```
JonasVanHoof, line 9: In the backend I have set this to `prefLabel` but for some treason it is using `preflabel` with a lower `L`. this is why im using the getter for the label here. => when the backend changes this is the place to change it and not in the "whole code base"

app/components/concept-scheme-uri-selector.js
JonasVanHoof: I tried it with the `sort: 'preflabel`' in the query but this is resulting in inconsisten results IMO 
image

PR #67, opened 2023-12-13
SFB-71: fix codelist options showing in preview

JonasVanHoof:
## ID
 [SFB-71](https://binnenland.atlassian.net/browse/SFB-71)

 ## Description

When adding a concept scheme to a field that is able to show options, these options should be visible. The information was not parsed into the meta graph of the store before which makes that showing the options was not possible. Because of PR https://github.com/lblod/app-form-builder/pull/7 this is possible.

Got all the information from the store and added it to it's own `concept-schemes.ttl` file. This file is than parsed in the preview store meta graph. Now the selectors are able to show the corresponding options.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Get a concept scheme configuration from the codelijsten tab. Copy the json and go back to the builder tab. In the builder tab create a field with a type that is able to show concept scheme options. for example the radio buttons or just the selector. Past the json configuration into the textarea field and see if all the options are visible in the preview form. 

 ## Links to other PR's

 - https://github.com/lblod/app-form-builder/pull/7
 - https://github.com/lblod/frontend-form-builder/pull/66


JonasVanHoof:
The data in the concept-schemes.ttl file is build up from the migrations that have been added to the backend app of the form builder. This is not ideal. 

I would have like it that we could fetch the store concept-schemes with their concepts and parse that into the graph but I couldn't find any option to do that without re assembling all the itemsreturned from the store to a triple that can be added to the ForkingStore graph

PR #70, opened 2023-12-18
added a remove button on the forms index

ZijneDwaasheid:
## ID
SFB-78

 ## Description

Now you have the option to delete forms in the index, this should take less clicks when wanting to delete a form.

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

### DIFF
--- a/app/controllers/index.js
+++ b/app/controllers/index.js
@@ -1,7 +1,42 @@
 import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
 
 export default class IndexController extends Controller {
+  @service store;
+  @service toaster;
+
   sort = '-created';
   page = 0;
   size = 20;
+
+  selectedForm;
+
+  @tracked showDeleteModal = false;
+
+  @action
+  OpenDeleteModal(generatedForm) {
+    this.selectedForm = generatedForm;
+    this.showDeleteModal = true;
+  }
+
+  @action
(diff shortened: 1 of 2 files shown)


app/templates/index.hbs
```diff
@@ -45,6 +45,16 @@
             <td class="au-u-visible-medium-up">
               <AuLink @route="formbuilder.edit" @model={{generatedForm.id}} @icon="pencil" @iconAlignment="left" @ariaHidden={{true}}>Bewerk</AuLink>
             </td>
+            <td>
```
Windvis, line 48: There is now an extra column without a corresponding `th`, which causes a visual bug with the column heading. Setting `colspan="3"` on line 38 should fix it.

app/controllers/index.js
```diff
@@ -1,7 +1,42 @@
...
+  @service store;
+  @service toaster;
+
   sort = '-created';
   page = 0;
   size = 20;
+
+  selectedForm;
+
+  @tracked showDeleteModal = false;
```
Windvis, line 16: Optional: it might be better to make the `selectedForm` tracked (and maybe rename it to `formToDelete`) and convert `showDeleteModal` to a getter that checks if a record is stored or not.

That way there is only 1 field that needs to be managed instead of 2.

Windvis (commented):
Some small suggestions to the code, but it works as expected 👍

Windvis (approved):

PR #71, opened 2023-12-27
SFB-86: (V2 Form) Sections not displayed in builder tab

ZijneDwaasheid:
… form:include

## ID
 SFB-86

 ## Description

Changing the formbuilder so we can see V2 sections in the formbuilder and add them.

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [x] Other

 ## How to test


ZijneDwaasheid:
Also changed validations to V2

PR #68, opened 2023-12-13
SFB-73: Optimise codelist data flow (DEPRECATED)

JonasVanHoof:
## ID
 [SFB-73](https://binnenland.atlassian.net/browse/SFB-73)

 ## Description

 Create a conditional dropdown for selecting the codelist. When the input field type is a `conceptSchemeMultiSelector`, `conceptSchemeRadioButtons`, `conceptSchemeSelector`. The codelist `concept-schemes uri selector` dropdown should be shown. When selecting a `concept-scheme` with that dropdown the fields `form:options` is added to the field subject.

**Field options**
```
form:options
        """{"conceptScheme":"http://data.vlaanderen.be/id/conceptscheme/BestuurseenheidClassificatieCode","searchEnabled":true}""";
```

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Select a input field type that is one of above and add a codelist to it. See if it is visualized in the preview screen.(visualization is only possible after PR: https://github.com/lblod/frontend-form-builder/pull/67 is merged)

 ## Links to other PR's

 - /

PR #72, opened 2024-01-03
SFB-73/optimize codelijsten 

JonasVanHoof:
## ID
 [SFB-73](https://binnenland.atlassian.net/browse/SFB-73)

 ## Description

The users do not understand how they have to add options to there dropdown component. They have to add a json configuration in a textarea to apply options for that dropdown.

I added a new tab in the topbar which is called `configuration` the idea behind this is that you can find all the extra configuration on sections/fields. Now I added a selector for the concept-scheme-uri's if the displaytype of the field is a selector component. They are listed in the code which ones.

With the idea of adding layout to fields/sections this can be added in this tab to.

Adding the previous branch here just as a reference for the code https://github.com/lblod/frontend-form-builder/pull/68

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Build a small form with minimum two sections (this is just for display) and a field that is a selector (eg. radiobuttons,selector)  in the builder. Once you did this you can add extra configuration to the fields that can hold `form:options`. Switch between the options for a field and look if the previous options are correctly applied and the old onces replaced. The preview should also refresh just as the code is updated in when switching to the code tab.

 ## Links to other PR's

- /


app/templates/formbuilder/edit/configuration.hbs
```diff
@@ -0,0 +1,56 @@
...
+            </c.content>
+          </AuCard>
+        {{/each}}
+      {{else}}
+        <span class="formBuilderEdit__noFields">
+          <p class="au-u-h3">No section selected</p>
+        </span>
+      {{/if}}
+    </div>
+  {{/if}}
```
JonasVanHoof, line 55: I think it is possible do this in an RDF-form

ZijneDwaasheid (commented):
If I go to the config tab and add a conceptscheme, then leave the tab and reenter the tab and selecting the same field. The selected conceptscheme is not selected anymore (in this tab, in the ttl code it stays, so it does what it needs to )

app/components/concept-scheme-uri-selector.js
```diff
@@ -11,14 +11,50 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
...
+    if (uri) {
+      const conceptScheme = await this.getconceptSchemeByUri(uri);
+
+      this.setSelected({
+        uri: uri,
+        label: conceptScheme ? conceptScheme.label : uri,
+      });
+    }
+  }
+
```
JonasVanHoof, line 48: @ZijneDwaasheid  this resolves the issue that the already applied concept scheme is not showing in the dropdown when going to the configuration view

ZijneDwaasheid (commented):
something else I just encountered, so you can select a section in the dropdown and then you see all the fields beneath it. This somehow not always the case. I add 3 fields and only see 2. 4 fields, also the first 2 fields. Only get max 2 fields.
Screenshot from 2024-01-16 10-08-36

ZijneDwaasheid:
> something else I just encountered, so you can select a section in the dropdown and then you see all the fields beneath it. This somehow not always the case. I add 3 fields and only see 2. 4 fields, also the first 2 fields. Only get max 2 fields. Screenshot from 2024-01-16 10-08-36

also when using fields where you can actually add a conceptscheme btw

app/components/concept-scheme-uri-selector.js
ZijneDwaasheid: this is now ok :)

JonasVanHoof:
> something else I just encountered, so you can select a section in the dropdown and then you see all the fields beneath it. This somehow not always the case. I add 3 fields and only see 2. 4 fields, also the first 2 fields. Only get max 2 fields. Screenshot from 2024-01-16 10-08-36

Good one! The fields array was defined to quickly what results in not showing all the fields in the view. 

This should be resolved. Tried adding multiple sections with different types of fields and everything seems ok now 😌  So hopefully you don't find any bugs XD

ZijneDwaasheid (approved):
Looks good to me now. Also like the work flow here. Choose a section and edit the fields linked to them.

PR #74, opened 2024-01-05
SFB-113: Clean up of duplicate validations in the code

JonasVanHoof:
## ID
 [SFB-113](https://binnenland.atlassian.net/browse/SFB-113)

 ## Description

Whit the latest refactor of the validations it is possible to add duplicate validations to a field. This is only possible when adding the duplicate validation right behind each other. This means that you do not leave the `validations` tab when doing this. When you switch tabs and try to add a duplicate the components are aware and than this is not possible anymore.

In this PR I created a method that will clean up any duplicate validations from the ttl code. I also tested this method so we can be more sure that this is working as expected.

This PR does not fix the possibility to add a duplicate validation in the same active tab!

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Go to the validations tab and add two validations of the same type
2. When going back to the code there should only be one 
3. There should be a console log that says that a duplicate validation is removed
4. When going back to the validations tab there will only be one validation
5. When wanting to add that duplicate again it refuses to

As we are removing items from the ttl code we have to be sure that it does not remove anything that it isn't supposed to so I also added some tests that will check if the method does not do anything strange.

 ## Links to other PR's

- /

### DIFF
--- a/app/components/add-validations-to-form.js
+++ b/app/components/add-validations-to-form.js
@@ -16,12 +16,14 @@ import { addValidationTriplesToFormNodesL } from '../utils/validation/add-field-
 import { getFieldAndValidationTriples } from '../utils/get-field-and-validation-triples';
 import { areValidationsInGraphValidated } from '../utils/validation/are-validations-in-graph-validated';
 import { createFieldDataForSubject } from '../utils/create-field-data-for-subject';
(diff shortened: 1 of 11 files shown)


ZijneDwaasheid (approved):
Looks good to me, does what it's expected to do.

PR #76, opened 2024-01-08
SFB-68 improving the error message of validations

ZijneDwaasheid:
## ID
 SFB-68 

 ## Description

 Adding default error messages to the validation tab

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

### DIFF
--- a/public/forms/validation/form.ttl
+++ b/public/forms/validation/form.ttl
@@ -164,7 +164,7 @@ fields:phoneNumberCountryCodeCFG a form:ConditionalFieldGroup ;
       sh:name "Fout melding" ;
       sh:order 100 ; # I think we always want to show this below the conditional fields?
       form:help """
-        Click here to see default error message or edit error message.
+        Klik hier om het standaardbericht te tonen of aan te passen.
       """ ;
       sh:path sh:resultMessage ;
       form:displayType displayTypes:defaultInput ;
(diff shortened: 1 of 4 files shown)


JonasVanHoof (changes requested):
Convinient!

The problem with first clicking in the input field to get the default message is that the user does not know that.. we have to hint the user that they can do that or something else that this is clear.

For example i had a form where i added an empty result message and then clicked in the input field what resulted in nothing. As it is already set.

When adding a validation and setting type REQUIRED and filling in the message. Than changing the type to MAX CHARACTERS and clicking in the result message input  wil toggle between the two messages instead of just the latest one. Added a mp4 file for explanation

* adding a check if the result message on the validation is empty or undefined
* Not sure how to fix the double result messages..

The code looks good but add those extra checks. If you find a  solution for that clicking thing let me know! Else this is something that will have to wait until [SFB-119](https://binnenland.atlassian.net/browse/SFB-119) is implemented

public/forms/validation/form.ttl
```diff
@@ -163,6 +163,9 @@ fields:phoneNumberCountryCodeCFG a form:ConditionalFieldGroup ;
     ext:validationResultMessageF a form:Field ;
       sh:name "Error message" ;
       sh:order 100 ; # I think we always want to show this below the conditional fields?
+      form:help """
+        Click here to see default error message or edit error message.
+      """ ;
```
JonasVanHoof, line 168: Will do the trick I think :)

JonasVanHoof (approved):

PR #77, opened 2024-01-08
SFB-116: add pencil icon next to title that opens the edit form name

JonasVanHoof:
## ID
 [SFB-116](https://binnenland.atlassian.net/browse/SFB-116)

 ## Description

The action to edit the name of a form is hidden behind the `formulier acties` dropdown. Add a pencil icon next to the Title and let it open the edit name and description modal.

 ## Type of change

 - [x] Small change
 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Click the `formulier acties` dropdown => the option to edit should be removed
2. Click on the pencil icon next to the title of the form and see if the modal opens
3. Change the name see if it updates

 ## Links to other PR's

- /

### DIFF
--- a/app/components/toolbar.js
+++ b/app/components/toolbar.js
@@ -77,4 +77,11 @@ export default class ToolbarComponent extends Component {
 
     this.args.setFormChanged(false);
   }
+
+  @action
+  closeEditNameModal() {
+    this.showEditModal = false;
+    this.formLabel = this.args.model.label;
+    this.formComment = this.args.model.comment;
+  }
 }
--- a/app/components/toolbar.hbs
+++ b/app/components/toolbar.hbs
@@ -83,7 +83,7 @@
 <AuModal
   @title="Bewerk naam en beschrijving"
   @modalOpen={{this.showEditModal}}
-  @closeModal={{fn (mut this.showEditModal) false}}
+  @closeModal={{this.closeEditNameModal}}
   as |Modal|
 >
   <Modal.Body>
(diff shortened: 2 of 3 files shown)


app/components/toolbar.js
```diff
@@ -77,4 +77,11 @@ export default class ToolbarComponent extends Component {
 
     this.args.setFormChanged(false);
   }
+
+  @action
+  closeEditNameModal() {
+    this.showEditModal = false;
+    this.formLabel = this.args.model.label;
+    this.formComment = this.args.model.comment;
```
JonasVanHoof, line 85: This is a bug fix*

ZijneDwaasheid (commented):
Maybe also add unique name and required name here

ZijneDwaasheid:
also, not that this has to be done here, the dropdown now holds 2 items. Maybe just put 'exporteer code' an 'verwijder' next to each other?

JonasVanHoof:
> Maybe also add unique name and required name here, you can now change the name to empty or a duplicate.

Should definitely be done! Not sure if I want to do it here. Also in the test document(which I apparently cannot open) was said that we wanted to remove the description from the form as it is never used.

That would also mean that editing the name of the form can just happen without a modal :)

JonasVanHoof:
> also, not that this has to be done here, the dropdown now holds 2 items. Maybe just put 'exporteer code' an 'verwijder' next to each other?

Good catch! I'm aware but not sure how we want to do it because next to each other is not that clean. Maybe change the export and delete to just be an icon with a message on hover?

JonasVanHoof:
> also, not that this has to be done here, the dropdown now holds 2 items. Maybe just put 'exporteer code' an 'verwijder' next to each other?

I have an idea on how to fix it but will do this in a different PR

PR #80, opened 2024-01-08
Refactor: forking store helper methods refactor

JonasVanHoof:
The Forking store helper methods are now exported one by one. I looked at another project and saw that the helper methods are exported as one object. This means that when importing the helper all the methods are also imported. This was not the case before as they were exported separately.

For development, the object is much easier as you can see what methods are available on the object. For me, this is a 💡  to refactor it.

Also removed two methods that were not used anymore from the forking-store-helpers utlis


Windvis:
> This means that when importing the helper all the methods are also imported.

This also means they are always bundled, even if they aren't used. If they are separate exports the bundler can tree-shake them out.

> For development, the object is much easier as you can see what methods are available on the object.

I think VS Code should be able to display all the available named exports as well? It sounds like a tooling issue to me, but I haven't looked into this myself.

Anyway, I really don't like this change. I see little value in prefixing all helpers. It just makes the code more verbose 😄. Just my opinion, of course. If everyone else on the team prefers this style, go for it.

PR #81, opened 2024-01-10
SFB-113: UI prevent duplicate validations on field

JonasVanHoof:
## ID
 [SFB-113](https://binnenland.atlassian.net/browse/SFB-113)

 ## Description

Whit the latest refactor of the validations it is possible to add duplicate validations to a field. This is only possible when adding the duplicate validation right behind each other. This means that you do not leave the `validations` tab when doing this. When you switch tabs and try to add a duplicate the components are aware and than this is not possible anymore.

In this PR i updated the form but it has to be re-rendered in the UI. This means that you will see the form re-loading when changing something in the form which is not ideal. I also added a warning notification above the validations that says that this is going to happen so the user knows that we are aware of this.

I haven't find a solution yet but we should be able to re-render the form within itself 💡 

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Go tot the validations tab  and select a field
2. Add a validation to the field
3. The form re-renders and is applied
4. When wanting to add a duplicate validation in  that same "session" this is not possible anymore

 ## Links to other PR's

 - https://github.com/lblod/frontend-form-builder/pull/74

### DIFF
--- a/app/components/add-validations-to-form.hbs
+++ b/app/components/add-validations-to-form.hbs
@@ -20,18 +20,36 @@
 
     <AuHr />
 
-    {{#if this.selectedField}}
-      <FieldValidationsForm
-        @fieldSubject={{this.selectedField.subject}}
-        @form={{this.selectedField.form}}
-        @graphs={{this.graphs}}
-        @store={{this.selectedField.store}}
-        @updateTtlCodeWithField={{this.updateTtlCodeWithField.perform}}
-      />
-    {{else}}
-      <span class="formBuilderEdit__noFields">
-        <p class="au-u-h3">No field selected</p>
-      </span>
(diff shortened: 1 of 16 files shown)


JonasVanHoof:
**Commit**
https://github.com/lblod/frontend-form-builder/pull/81/commits/39f0052c00b72d3d583e617fab166b37c8e77d3c

Not sure if it makes it better but now you do not see that the form is loaded in from top to bottom. It of course always needs some time to show in the UI so the animation time cannot go lower unfortunately.

JonasVanHoof:
Closed as we better not merge temporary branches

PR #87, opened 2024-01-17
Serialize ember data model to turtle

JonasVanHoof:
All concept schemes that can be used in the form-builder come from the database but for the preview, these concept-schemes and concepts are stored in a static file. This means that when creating new concept schemes we do not have the latest data to show in our preview which results in an empty list and confusion for our users.

In this PR I created a service that will get all the latest concept schemes with their concepts from the database when going to the form overview page. If the service returns null for the TTL of the concept-schemes it will use the local file in the form-builder public folder.


JonasVanHoof:
Do not review this as we are postponing this functionality

cc: @aatauil @ZijneDwaasheid

PR #88, opened 2024-01-19
refactor: forkingstore helpers

JonasVanHoof:
Some methods in the helper where not used and some could be moved to the parent component so that's what I did :D


JonasVanHoof:
Found this bug when refactoring the forkingstore helpers. This has nothing to do with the code I refactored but just letting yyou know i stumbled on this here https://binnenland.atlassian.net/browse/SFB-121

PR #90, opened 2024-01-22
Refactor: cleanup the namespaces (utils/rdflib)

JonasVanHoof:
## Issue
The namespaces are  defined in the `ember-submission-form-field-helpers` addon. We are redefining them in the form-builder. 

## Fix
Use the namespaces from the `ember-submission-form-field-helpers`

## Test
- All functionality


app/components/field-validations-form.js
```diff
@@ -3,7 +3,8 @@ import { tracked } from '@glimmer/tracking';
 import { restartableTask, timeout } from 'ember-concurrency';
 import { getFieldAndValidationTriples } from '../utils/get-field-and-validation-triples';
 import { getValidationSubjectsOnNode } from '../utils/forking-store-helpers';
-import { EXT, FORM } from '../utils/rdflib';
+import { EXT } from '../utils/namespaces';
```
JonasVanHoof, line 6: Renamed our _rdflib.js_ file to _namespaces.js_. This is how it is called in the `ember-submission-form-field-helpers`. Our namespaces are extending the ones defined in the helper addon.

JonasVanHoof:
I cleaned up the namespaces and only one is in there now. EXT is exported from our namespaces.js file which is ok because it is not exported from the helpers. The EXT namespaces are used in the helpers but they use specific namespaces e.g
```js
const volunteersPredicate = namedNode(
  "http://mu.semte.ch/vocabularies/ext/volunteers"
);
```

With that code in mind, I was thinking of defining our Predicate namespaces like `form:validations` or `sh:group` in this file. All base namespaces will be only imported in this file and the namespaces used in the frontend files will be imported from the app itself. If something changes we only have to extend our file of namespaces.

What do you think?

Tested this in our code with the following: 
```js
export const SH_GROUP = SHACL('group');
```

@aatauil  @ZijneDwaasheid

aatauil (approved):

JonasVanHoof:
> With that code in mind, I was thinking of defining our Predicate namespaces like `form:validations` or `sh:group` in this file. All base namespaces will be only imported in this file and the namespaces used in the frontend files will be imported from the app itself. If something changes we only have to extend our file of namespaces.
> 
> What do you think?
> 


We will not be creating the specific namepsaces like below. cc: @aatauil  e.g.
> ```js
> export const SH_GROUP = SHACL('group');
> ```

PR #91, opened 2024-01-23
Refactor: duplicate methods for sorting

JonasVanHoof:
## Issue and fix

In many files  a sorting method is created at the top of the file. This method was the same for a lot of files. Since we created a `sort objects on property` util method we can use that one so we do not have to redefine the `byLabel` method in each file.

```js
function byLabel(a, b) {
  const textA = a.label.toUpperCase();
  const textB = b.label.toUpperCase();
  return textA < textB ? -1 : textA > textB ? 1 : 0;
}
```

## Test
1. all items should be sorted ascernding
2. except for the validation type options I hyave set these on descending so the "Verplicht" options what is used the most is at the top of the list
3. Created a basic unit test that will show that we can sort on property `label` _ascending_ and _descending_.

### DIFF
--- a/app/components/rdf-form-fields/validation-concept-scheme-selector.js
+++ b/app/components/rdf-form-fields/validation-concept-scheme-selector.js
@@ -85,7 +85,8 @@ export default class ValidationConceptSchemeSelectorComponent extends InputField
     });
     this.validationTypeOptions = sortObjectsOnProperty(
       filteredOptions,
-      'label'
+      'label',
+      false
     );
   }
 
--- a/app/components/rdf-form-fields/validation-concept-scheme-selector.js
+++ b/app/components/rdf-form-fields/validation-concept-scheme-selector.js
@@ -17,12 +17,7 @@ import {
 import { showErrorToasterMessage } from '../../utils/toaster-message-helper';
 import { getGroupingTypeForValidation } from '../../utils/validation/get-grouping-type-for-validation';
 import { getDefaultErrorMessageForValidation } from '../../utils/validation/get-default-error-message-for-validation';
-
-function byLabel(a, b) {
-  const textA = a.label.toUpperCase();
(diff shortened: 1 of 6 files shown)


app/components/rdf-form-fields/validation-concept-scheme-selector.js
```diff
@@ -84,12 +79,16 @@ export default class ValidationConceptSchemeSelectorComponent extends InputField
...
       return conceptOptions
         .map((concept) => concept.value)
         .includes(option.subject.value);
     });
-    this.validationTypeOptions.sort(byLabel);
+    this.validationTypeOptions = sortObjectsOnProperty(
+      filteredOptions,
+      'label',
+      false
+    );
```
JonasVanHoof, line 91: Validations sorted descending

app/utils/sort-object-on-property.js
```diff
@@ -1,4 +1,8 @@
-export function sortObjectsOnProperty(arrayOfObjects, property) {
+export function sortObjectsOnProperty(
+  arrayOfObjects,
+  property,
+  sortAscending = true
```
JonasVanHoof, line 4: Should we change it to ASC & DESC?

tests/unit/utils/sort-object-on-property-test.js
```diff
@@ -0,0 +1,92 @@
+import { sortObjectsOnProperty } from 'frontend-form-builder/utils/sort-object-on-property';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | sortObjectsOnProperty', function () {
+  module('can sort', function () {
+    test('ascending on property label', function (assert) {
```
JonasVanHoof, line 6: Only tested on property labe;. Maybe it is good to also test this with numbers and date's
(Sort on numbers is used in the code already and is visually working)

PR #92, opened 2024-01-23
feat: add feature flag for create codelist button

JonasVanHoof:
## Issue

Users can make their own codelists but those itemst will not be visible in the dropdown when they want to assign a codelist to a field. 

## Fix

Put the create button behind a feature flag `CAN_CREATE_OWN_CODELIST` and set it to `false`.

## Test

- Can you create a new codelists?


config/environment.js
```diff
@@ -20,6 +20,7 @@ module.exports = function (environment) {
     },
     featureFlags: {
       SHOW_ORIGNAL_CODELIST_PAGE: false,
+      CAN_CREATE_OWN_CODELIST: false,
```
ZijneDwaasheid, line 23: maybe mention here, that you commented new route in the router.js

PR #96, opened 2024-01-24
SFB-122:double error messages

ZijneDwaasheid:
## ID
 SFB-122

 ## Description

It is possible to add multiple error messages to a validation, this shouldn't be possible. Fixed the double error messages by making a custom rdf form field. This input field will remove the previous error message when making a new one.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test
1) choose a validation type
2) Click in the input field, the default error message will show (if there is one of course) or type a custom one
3) try changing validation type
4) if you want the default error message of that validation type remove the old one and just click out of the input field or again type custom one

 ## Links to other PR's

### DIFF
--- a/app/components/rdf-form-fields/error-message-input-field.js
+++ b/app/components/rdf-form-fields/error-message-input-field.js
@@ -28,9 +28,9 @@ export default class ErrorMessageInputFieldComponent extends SimpleInputFieldCom
         this.value,
         this.storeOptions.sourceGraph
       );
-      this.removeOldMessage();
       this.args.formStore.addAll([messageToForm]);
     }
+    this.removeOldMessage();
   }
 
   @action
--- a/app/components/rdf-form-fields/error-message-input-field.js
+++ b/app/components/rdf-form-fields/error-message-input-field.js
@@ -2,13 +2,16 @@ import SimpleInputFieldComponent from '@lblod/ember-submission-form-fields/compo
 import { action } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
 import { tracked } from '@glimmer/tracking';
-import { SHACL } from '@lblod/submission-form-helpers';
+import { SHACL, RDF } from '@lblod/submission-form-helpers';
 import { Statement } from 'rdflib';
+import { getDefaultErrorMessageForValidation } from '../../utils/validation/get-default-error-message-for-validation';
+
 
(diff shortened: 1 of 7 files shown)


app/components/rdf-form-fields/error-message-input-field.js
```diff
@@ -0,0 +1,54 @@
...
+      const errorMessageInForm = this.args.formStore.any(
+        this.storeOptions.sourceNode,
+        SHACL('resultMessage'),
+        undefined,
+        this.storeOptions.sourceGraph
+      );
+      if (errorMessageInForm) {
+        this.value = errorMessageInForm;
+      }
+    }
```
JonasVanHoof, line 42: If I write a result message:
1. focus out of input field
2. clear the input field as i dont want it
3. Focus out it of the field
4. It sets back the value that was before

Think that the the else is wrong here

JonasVanHoof (changes requested):

JonasVanHoof (changes requested):
This makes it easier for the user to add or remove the default error message. A remark: When toggling the switch to use the default error messsage it sets the message from the database, that's ok. When going away and coming back the toggle is not active anymore and the input field is switched to a custom input. When removing the content in the input field and going back to the code it did not remove it. You can tell by the ttl code and also that the value is filled in again when switching to the validation.

I would also make it that the default message is added by default and if you want a custom one that it shows the input field when toggled off.

app/components/rdf-form-fields/error-message-input-field.js
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  get toggleMessage() {
+    if (this.useDefaultMessage) {
+      return 'Toggle om een aangepaste foutmelding te gebruiken.';
+    } else {
+      return 'Toggle om een standaard foutmelding te gebruiken.';
+    }
+  }
```
JonasVanHoof, line 119: Maybe overkill 
```js

  get toggleMessage() {
    const showMessage = (change) =>
      `Toggle om een ${change} foutmelding te gebruiken.`;

    if (this.useDefaultMessage) {
      return showMessage(`aangepaste`);
    } else {
      return showMessage(`standaard`);
    }
  }
```

JonasVanHoof (approved):

PR #98, opened 2024-01-29
SFB-137: When routing away from the edit route show save modal

JonasVanHoof:
## ID
 [SFB-137](https://binnenland.atlassian.net/browse/SFB-137)

 ## Description

When routing away from a `formbuilder/edit` route when the user has unsaved changes show a modal where the user can save or discard it's changes.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. open a form
2. Route to `codelijsten` ( top right)
3. did you see a modal? NO
4. Go back to your form and change the title of the section
5. Go to codelijsten
6. Did you see a modal? YES
7. Try the discard first and than the save unsaved changes
 
## Links to other PR's

 - /

### DIFF
--- a/app/controllers/formbuilder/edit.js
+++ b/app/controllers/formbuilder/edit.js
@@ -107,7 +107,7 @@ export default class FormbuilderEditController extends Controller {
   @action
   saveUnsavedChanges() {
     this.formCodeManager.addFormCode(
-      this.formCodeManager.getTtlOfLatestVersion
+      this.formCodeManager.getTtlOfLatestVersion()
     );
     this.formCodeManager.pinLatestVersionAsReference();
     this.isSaveModalOpen = false;
@@ -118,7 +118,7 @@ export default class FormbuilderEditController extends Controller {
   discardSave() {
     this.isSaveModalOpen = false;
     this.formCodeManager.addFormCode(
-      this.formCodeManager.getTtlOfLatestVersion()
+      this.formCodeManager.getTtlOfReferenceVersion()
     );
     this.goToNextRoute();
   }
--- a/app/controllers/formbuilder/edit.js
+++ b/app/controllers/formbuilder/edit.js
@@ -31,6 +31,9 @@ export default class FormbuilderEditController extends Controller {
   @tracked previewStore;
   @tracked previewForm;
 
+  @tracked isSaveModalOpen;
+  @tracked nextRoute;
+
   sourceNode = PREVIEW_SOURCE_NODE;
 
   @action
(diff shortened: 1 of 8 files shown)


JonasVanHoof:
**NOTE:** When routing away from a page without using the router the modal will not be shown. @aatauil is this something we should handle or is this to intentionally?

JonasVanHoof:
@Windvis  Did you already implement somehting like this in LDPC project? [ticket ref](https://binnenland.atlassian.net/browse/SFB-137?focusedCommentId=96450)

The idea behind it is that when routing away from a page where data can be changed we want to show a modal that will give the user the option to save or discard it's changes. In the current implementation in this PR this route change is only detected when routing with the `this.router`  with the router hook `willTransition`.

Do you have a link to your implementation or any idea what router hook can be used to also detect these changes?

Windvis:
Yea, there are some variations across the different projects. The one in [LPDC](https://github.com/lblod/frontend-lpdc/commit/72f1728d77ff861273237c3bd0e9b36f40b58068) uses the `willTransition` _event_ and sets it up in a component. This is the [commit](https://github.com/lblod/frontend-lpdc/commit/72f1728d77ff861273237c3bd0e9b36f40b58068), but there are probably follow up commits. The latest version can be found in the frontend-lpdc repo. (Loket removed the LPDC module).

I think we also have a "generic" solution somewhere where you add a component which does the setup and shows the modal based on a boolean argument. Not sure where we used that, but I remember building something at least 🤔.

In any case, there are multiple ways of doing things. Are there cases where this PR doesn't work?

Windvis:
Oh, wait I just saw this comment:

> NOTE: When routing away from a page without using the router the modal will not be shown.

What does this mean exactly? There is not really a way to route without the router? Or do you mean manual URL edits / refreshes?

JonasVanHoof:
> Oh, wait I just saw this comment:
> 
> > NOTE: When routing away from a page without using the router the modal will not be shown.
> 
> What does this mean exactly? There is not really a way to route without the router? Or do you mean manual URL edits / refreshes?

Yes indeed! The usecase that work are:
* go back in the breadcrums bar (ember router service)
*  Go to a link `codelijsten` for example (ember router service)
* Pressing the back button in the browser (native)

The only thing I found that was not working is when manually changing the url (to e.g. the index page ) or refreshing the page that the popup is not shown. Is there a solution for this to block it?

JonasVanHoof:
> Member

Hehe, I don't think it is a hard requirement  Achraf  suggested to connect with you for this as you already did something like this and that we are sure that this is not something you already fixed :D  But if your version is doing the same as mine it is solved I think? @aatauil

aatauil (approved):
Made some small UX changes, other then that, looking good :)

PR #101, opened 2024-01-29
change ESFF version that uses V2 model

ZijneDwaasheid:
## ID
 SFB-131

 ## Description
This uses the newer version of ESFF that uses the newer V2 form model

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test
For now you can only preview V2 forms.

 ## Links to other PR's

### DIFF
--- a/app/components/rdf-form-fields/property-group-selector.js
+++ b/app/components/rdf-form-fields/property-group-selector.js
@@ -27,7 +27,7 @@ export default class PropertyGroupSelector extends InputFieldComponent {
   loadOptions() {
     const sourceGraph = this.args.graphs.sourceGraph;
     const propertyGroup = new namedNode(
-      'http://lblod.data.gift/vocabularies/forms/PropertyGroup'
+      'http://lblod.data.gift/vocabularies/forms/Section'
     );
 
     const allRawOptions = this.args.formStore
--- a/public/forms/builder/form.ttl
+++ b/public/forms/builder/form.ttl
@@ -17,7 +17,7 @@ ext:form a form:Form, form:TopLevelForm ;
   form:includes ext:formListingTableL.
 
 ##########################################################
-#  properform:partOf
+#  Section
 ##########################################################
 ext:mainPg a form:Section;
     sh:name "Form Builder" ;
@@ -237,7 +237,7 @@ ext:formListingFormItem a form:SubForm;
   form:includes ext:formListingAddF;
   form:includes ext:formListingRemoveF;
   form:includes ext:subformNodesL;
-  form:includes ext:subFormsectionF;
+  form:includes ext:subFormSectionF;
   form:hasFieldGroup ext:listingPg.
 
 ext:formListingAddF a form:Field ;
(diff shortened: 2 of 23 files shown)


JonasVanHoof:
It's latest so you could not have done this yet but there is a new version ;p https://github.com/lblod/ember-submission-form-fields/commit/c9563fb92b60b24d58ab90afcab03f14d7ce5777

public/forms/builder/form.ttl
```diff
@@ -11,96 +11,96 @@
...
-#  property-group
+#  Section
 ##########################################################
-ext:mainPg a form:PropertyGroup;
+ext:mainPg a form:Section;
     sh:name "Form Builder" ;
     sh:order 0 .
 
-ext:propertyGroupPg a form:PropertyGroup;
+ext:sectionPg a form:Section;
```
JonasVanHoof, line 26: `ext:SectionPg` is not correct anymore as `Pg` stands for `PropertyGroup` The same for the subjects below.

public/forms/builder/form.ttl
```diff
@@ -11,96 +11,96 @@
...
 #    with no "sh:path" provided.
-#    We now revert to attaching the propertyGroups to
+#    We now revert to attaching the sections to
 #    the "form:Form" directly.
 #    Which is suboptimal, cause not really in the defined
 #    model. (It's not forbidden though)
 ##########################################################
-ext:propertyGroupS a form:Scope;
-  sh:path sh:group ;
+ext:sectionS a form:Scope;
```
JonasVanHoof, line 47: Than we have to find something for scope but my SC is good enough? `ext:sectionSC` I would even be happy with `ext:sectionScope`

public/forms/validation/form.ttl
```diff
@@ -19,7 +19,7 @@ ext:form a form:Form, form:TopLevelForm ;
 ##########################################################
 #  Property-groups
 ##########################################################
-ext:formFieldPg a form:PropertyGroup;
+ext:formFieldPg a form:Section;
```
JonasVanHoof, line 22: Also here `ext:formFieldPG` incorrect

JonasVanHoof (changes requested):
I think we should update the naming. As refactoring it to a V2 form improves the readability of the ttl code there cannnot be naming in the ttl that referneces to the V1 form.
* Pg = propertyGroup => S  = Section
* S = Scope => may7be just Scope as it is a very short word

app/components/rdf-form-fields/property-group-selector.js
```diff
@@ -27,7 +27,7 @@ export default class PropertyGroupSelector extends InputFieldComponent {
   loadOptions() {
     const sourceGraph = this.args.graphs.sourceGraph;
     const propertyGroup = new namedNode(
-      'http://lblod.data.gift/vocabularies/forms/PropertyGroup'
+      'http://lblod.data.gift/vocabularies/forms/Section'
     );
```
JonasVanHoof, line 31: This keeps annoying me haha I know there is a ticket for it 👌🏻 
Don't forget to also rename the component in that ticket :)

JonasVanHoof (commented):
The JSON configuration in the builder is back. Why is this? As the configuration tab is already in place I assume that something went wrong while rebasing. Can you check?

image

JonasVanHoof (approved):

PR #105, opened 2024-01-29
SFB147: v2 cant edit validations in validation tab

ZijneDwaasheid:
## ID
 SFB147

 ## Description
After updating to V2 forms we need the validations tab to be updated

 ## Type of change

 - [x] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

### DIFF
--- a/app/components/rdf-form-fields/property-group-selector.js
+++ b/app/components/rdf-form-fields/property-group-selector.js
@@ -27,7 +27,7 @@ export default class PropertyGroupSelector extends InputFieldComponent {
   loadOptions() {
     const sourceGraph = this.args.graphs.sourceGraph;
     const propertyGroup = new namedNode(
-      'http://lblod.data.gift/vocabularies/forms/PropertyGroup'
+      'http://lblod.data.gift/vocabularies/forms/Section'
     );
 
     const allRawOptions = this.args.formStore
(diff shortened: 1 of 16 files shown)


app/components/rdf-form-fields/property-group-selector.js
```diff
@@ -27,7 +27,7 @@ export default class PropertyGroupSelector extends InputFieldComponent {
   loadOptions() {
     const sourceGraph = this.args.graphs.sourceGraph;
     const propertyGroup = new namedNode(
```
JonasVanHoof, line 29: `const section = `

JonasVanHoof, line 29: const section =

app/utils/ttl-templates/basic-form-template.js
```diff
@@ -8,14 +8,14 @@ export default `@prefix : <#>.
 nodes:24289e48-258f-4919-8c3e-5783a6acb4a4
     a form:Field;
```
JonasVanHoof, line 9: Maybe for the ttl files change the base to the other branch where you already update this 😅 Otherwise same comments

JonasVanHoof (commented):
Works as expected , good job!

See what you do with the comments but a change of base branch looks best to me

app/components/rdf-form-fields/property-group-selector.js
JonasVanHoof: https://binnenland.atlassian.net/browse/SFB-149

PR #111, opened 2024-02-02
SFB-154: fetch concept schemes from db

JonasVanHoof:
## ID
 [SFB-154](https://binnenland.atlassian.net/browse/SFB-154)

 ## Description

now that users can create their own codelists the preview cannot show all the latest.  All the concept-schemes used in the form-builder are put in a local file and fetched every time we route to the `formbuilder.edit` route. 

In this PR we look up what conceptschemes are used in the ttlcode and depending on that we build up the ttl code for the used conceptschemes.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

1. Concept -schemes are not fetched locally
2. The preview shows the latest concept-schemes with their options
3. Create a new concept-scheme save it and see if it is visible in the preview form

 ## Links to other PR's

 - /


app/controllers/formbuilder/edit.js
```diff
@@ -157,4 +161,59 @@ export default class FormbuilderEditController extends Controller {
...
+      const optionsAsString = triple.object;
+      try {
+        const jsonOptions = JSON.parse(optionsAsString);
+        conceptSchemeUris.push(jsonOptions.conceptScheme);
+      } catch (error) {
+        return;
+      }
+    }
+    return conceptSchemeUris;
+  }
```
JonasVanHoof, line 218: Maybe split this up into smaller util methods? The reason for this would be that this is logic for one specifi subject. When calling one or two methods in this controller it will be easier to read and understand what is happening?

app/models/concept-scheme.js
```diff
@@ -19,4 +20,29 @@ export default class ConceptSchemeModel extends Model {
...
     return this.createdat ?? null;
   }
+
+  asTtlCode() {
+    return `
+      <${this.uri}>
+      ${RDF('type')} ${SKOS('ConceptScheme')} ;
+      ${SKOS('prefLabel')} "${this.label}" .
+    `;
+  }
```
JonasVanHoof, line 30: This fits into the model context

app/models/concept-scheme.js
```diff
@@ -19,4 +20,29 @@ export default class ConceptSchemeModel extends Model {
...
+  async modelWithConceptsAsTtlCode() {
+    const schemeTtlCode = this.asTtlCode();
+    const ttlCodeArray = [];
+
+    for (const concept of await this.getConceptModels()) {
+      ttlCodeArray.push(concept.asTtlCode(this.uri));
+    }
+
+    return [...ttlCodeArray, schemeTtlCode].join(' ');
+  }
```
JonasVanHoof, line 41: Doubting if this fits in the model context of concept-scheme?

app/routes/formbuilder/edit.js
```diff
@@ -21,19 +21,16 @@ export default class FormbuilderEditRoute extends Route {
 
   async model(params) {
     this.formCodeManager.setFormId(params.id);
-    const [generatedForm, formTtl, metaTtl, conceptSchemesTtl] =
-      await Promise.all([
-        this.getGeneratedFormById(params.id),
-        getLocalFileContentAsText('/forms/builder/form.ttl'),
-        getLocalFileContentAsText('/forms/builder/meta.ttl'),
-        getLocalFileContentAsText('/forms/preview/concept-schemes.ttl'),
```
JonasVanHoof, line 29: We got rid of the local concept-scheme file!

JonasVanHoof (commented):
It works as expected. No lag atm. Possibly when having alot of concept schemes in the form it is possible that it get's slower. The loading dots will be visible when loading all the data so this is not a problem if it happens.

@aatauil want to check with you if we want to do it like this.

app/controllers/formbuilder/edit.js
aatauil: I dont think that is nescessary, if later on we do want to reuse some smaller piece of code, there we actually turn it unto a util. For now this is good :)

app/models/concept-scheme.js
aatauil: we will search for a backend solution later on. For now this will do

aatauil (approved):

PR #112, opened 2024-02-02
SFB-155 :awareness of field path

JonasVanHoof:
## ID
 [SFB-155](https://binnenland.atlassian.net/browse/SFB-155)

 ## Description

Show the path value of the field in the configuration tab.

**Default paths**
image

**Complex path**
image

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Is the specified path in the ttl code visible in the configuration tab.

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/concept-scheme-uri-selector.hbs
+++ b/app/components/concept-scheme-uri-selector.hbs
@@ -1,9 +1,11 @@
-<AuLabel>{{t "codelists.chooseConcept"}}</AuLabel>
-<PowerSelect
-  @options={{this.options}}
-  @selected={{this.selected}}
-  @onChange={{this.setSelected}}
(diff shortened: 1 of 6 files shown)


JonasVanHoof:
@aatauil Making it possible that a smart-user can create a complex path. Is not possible atm... If you have an idea on how we can tranform the paths into a input field and from that input field to the ttl code, shoot! 

As the complex path's look like blank nodes we could use a small code editor in the configuration tab but I don't see why the user could not fill it in in the code editor?

Just to make the user more aware (as this was also something we wanted to) of the sh:path  I added this. Not sure how to improve this. I need a more concreet idea on this.

PR #116, opened 2024-02-07
Sfb 125/extended disable tabs when v1 form

JonasVanHoof:
## ID
 SFB-125

 ## Description

Disbeld the edit tabs builder, validations and configuration when the form is a V1 form. As links cannot be disabled i added buttons that will be show when the form is a V1 form.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

See if the tabs are disabled when it is a v1 form. When pasting or changing it to a V2 form the buttons will be clickable again.

 ## Links to other PR's

 - http://github.com/PR/2314135

## Attachments

image


aatauil (approved):

aatauil:
Good stuff

PR #117, opened 2024-02-12
SFB-156: export custom codelist

JonasVanHoof:
## ID
 [SFB-156](https://binnenland.atlassian.net/browse/SFB-156)

 ## Description

When users make their own codelists they can only see this list in the codelist details page or in the preview form when creating a form. Of course they also want the information from the codelist in their application. We add an extra option to the split button in the detail view so users can export the code code as a ttl file.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. create/go to a custom codelist
2. Press the split button and donwload the file.
3. When opening the file the content of the codelist should be in there
4. As the toolbar also uses the same new download method try and export the ttl code of a form 

 ## Links to other PR's

 - /
 
## Attachments
image

### DIFF
--- a/app/components/toolbar.js
+++ b/app/components/toolbar.js
@@ -6,6 +6,7 @@ import {
   DESCRIPTION_NOT_USED_PLACEHOLDER,
   NAME_INPUT_CHAR_LIMIT,
 } from '../utils/constants';
+import { downloadTextAsFile } from '../utils/download-text-as-file';
 
 export default class ToolbarComponent extends Component {
   @service store;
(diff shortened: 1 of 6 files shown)


app/components/toolbar.js
```diff
@@ -21,17 +22,14 @@ export default class ToolbarComponent extends Component {
...
-    let blob = new Blob([this.args.code], { type: 'text/plain' });
-    downloadLink.href = window.URL.createObjectURL(blob);
-
-    // Click file to download then destroy link
-    downloadLink.click();
-    downloadLink.remove();
+    downloadTextAsFile(
+      {
+        filename: this.formLabel,
+        contentAsText: this.args.code,
```
JonasVanHoof, line 28: contentAsString?

app/utils/download-text-as-file.js
```diff
@@ -0,0 +1,17 @@
+export function downloadTextAsFile(
```
JonasVanHoof, line 1: Maybe create a promise for this if it ever takes a longer time

ZijneDwaasheid (approved):

PR #120, opened 2024-02-15
SFB-168:  Archive custom codelists

JonasVanHoof:
## ID
 [SFB-168](https://binnenland.atlassian.net/browse/SFB-168)

 ## Description

Users can create custom codelists now. As other users can also make use of these codelists it's better to archive the codelists that to delete it. This makes it possible to restore the codelist if we want to.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Create a codelist
2. Archive the codelist
3. You should see the archived codelist in the overview list but with a new tag "Ge-archiveerd"
4. When opening the archived list you will only be able to view the codelist

 ## Links to other PR's

 -  https://github.com/lblod/app-form-builder/pull/9

### DIFF
--- a/app/components/concept-scheme-uri-selector.js
+++ b/app/components/concept-scheme-uri-selector.js
@@ -56,7 +56,11 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
       },
     });
 
-    this.options = sortObjectsOnProperty([...conceptSchemes], 'label');
+    const notArchivedConceptSchemes = [...conceptSchemes].filter(
+      (scheme) => !scheme.isArchived
+    );
+
+    this.options = sortObjectsOnProperty(notArchivedConceptSchemes, 'label');
 
     if (this.args.forField) {
       await this.loadSelected();
--- a/app/templates/codelijsten/edit.hbs
+++ b/app/templates/codelijsten/edit.hbs
@@ -82,7 +82,7 @@
       </AuAlert>
     {{else}}
       {{#if this.isArchivedConceptScheme}}
-        <AuAlert @title={{"Opgelet"}} @skin={{"warning"}} @icon={{"login"}}>
+        <AuAlert @title={{"Opgelet"}} @skin={{"info"}} @icon={{"archive"}}>
           <p>{{t "codelists.alert.archivedCodelist"}}</p>
         </AuAlert>
       {{/if}}
--- a/app/templates/codelijsten/edit.hbs
+++ b/app/templates/codelijsten/edit.hbs
@@ -26,7 +26,7 @@
         {{t "crud.save"}}
       </AuButton>
 
-      {{#if this.isPrivateConceptScheme}}
+      {{#if this.isReadOnly}}
         <span id="saveSplitButton_dropdown">
           <AuButton
             @disabled={{true}}
@@ -80,7 +80,14 @@
       <AuAlert @title={{"Opgelet"}} @skin={{"warning"}} @icon={{"login"}}>
         <p>{{t "codelists.alert.genericCodelist"}}</p>
       </AuAlert>
+    {{else}}
+      {{#if this.isArchivedConceptScheme}}
+        <AuAlert @title={{"Opgelet"}} @skin={{"warning"}} @icon={{"login"}}>
+          <p>{{t "codelists.alert.archivedCodelist"}}</p>
+        </AuAlert>
+      {{/if}}
     {{/if}}
+
     <div class="au-u-3-5 au-u-margin-bottom-large">
       <AuLabel for="input-codelist-name" @required={{true}}>
         {{t "codelists.nameOfCodelist"}}
@@ -89,7 +96,7 @@
         id="input-codelist-name"
         @width="block"
         @error={{this.nameErrorMessage}}
-        @disabled={{this.isPrivateConceptScheme}}
+        @disabled={{this.isReadOnly}}
         value={{this.codelistName}}
         {{on "blur" this.handleNameChange}}
       />
@@ -98,7 +105,7 @@
       {{/if}}
     </div>
     <div class="au-u-3-5 au-u-margin-bottom-large">
-      {{#if this.isPrivateConceptScheme}}
+      {{#if this.isReadOnly}}
         <p>{{this.codelistDescription}}</p>
       {{else}}
         <AuLabel for="input-codelist-description" @required={{true}}>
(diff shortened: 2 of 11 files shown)


app/templates/codelijsten/edit.hbs
```diff
@@ -49,15 +49,15 @@
...
             @alert={{true}}
             role="menuitem"
             @disabled={{this.isPrivateConceptScheme}}
-            @loadingMessage={{t "messages.loading.isDeleting"}}
+            @loadingMessage={{t "messages.loading.isArchiving"}}
             {{on "click" (fn (mut this.isDeleteModalOpen) true)}}
           >
-            {{t "crud.delete"}}
+            {{t "crud.archive"}}
           </AuButton>
```
JonasVanHoof, line 61: What color should the archive button be? It's important to let the user know that this is something serious (red) but on the other hand it is only an archive so it can be blue. Maybe orange?

image

app/templates/codelijsten/edit.hbs
```diff
@@ -173,7 +173,10 @@
   </div>
 </div>
 
-<AuModal @modalOpen={{this.isDeleteModalOpen}} @closeModal={{this.close}}>
+<AuModal
+  @modalOpen={{this.isDeleteModalOpen}}
+  @closeModal={{fn (mut this.isDeleteModalOpen) false}}
+>
```
JonasVanHoof, line 179: BUG: when closing the modal with the `X` it did nothing

app/templates/codelijsten/index.hbs
```diff
@@ -66,6 +66,14 @@
...
               >{{t "codelists.types.isPublic"}}
               </AuPill>
+              {{#if conceptScheme.isArchived}}
+                <AuPill
+                  @skin="default"
+                  @icon="archive"
+                  @iconAlignment="left"
+                >{{t "codelists.types.isArchived"}}
+                </AuPill>
+              {{/if}}
```
JonasVanHoof, line 76: image

app/templates/codelijsten/edit.hbs
```diff
@@ -71,7 +70,14 @@
       <AuAlert @title={{"Opgelet"}} @skin={{"warning"}} @icon={{"login"}}>
         <p>{{t "codelists.alert.genericCodelist"}}</p>
       </AuAlert>
+    {{else}}
+      {{#if this.isArchivedConceptScheme}}
+        <AuAlert @title={{"Opgelet"}} @skin={{"warning"}} @icon={{"login"}}>
+          <p>{{t "codelists.alert.archivedCodelist"}}</p>
+        </AuAlert>
+      {{/if}}
```
JonasVanHoof, line 78: image

app/templates/codelijsten/edit.hbs
```diff
@@ -49,161 +51,174 @@
...
-        {{on "blur" this.handleNameChange}}
-      />
-      {{#if this.nameErrorMessage}}
-        <AuHelpText @error={{true}}>{{this.nameErrorMessage}}</AuHelpText>
-      {{/if}}
-    </div>
-    <div class="au-u-3-5 au-u-margin-bottom-large">
+{{#if this.setup.isRunning}}
+  <AuLoader />
+{{else}}
```
JonasVanHoof, line 81: Add a loader to the content so it is not flickering between previous states of the codelist detail

ZijneDwaasheid (commented):
Ok so tested this and followed the 'how to test', at step 3 You may not be able to see the codelists in the overview. So first of all it gets the tag archived, but then you see it of course still in the overview and you're still able to add the concept-scheme in the formbuilder. I don't know this what you expected or I'm missing something here.

JonasVanHoof:
> Ok so tested this and followed the 'how to test', at step 3 You may not be able to see the codelists in the overview. So first of all it gets the tag archived, but then you see it of course still in the overview and you're still able to add the concept-scheme in the formbuilder. I don't know this what you expected or I'm missing something here.

Ah im sorry didn't update the test description but this is indeed not the expected behaviour  that was written down. With the thing in mind how should we recover from an archived list I didn't remove it from the overview page and just added the tag. This makes it possible for the user to see the codelist and that they not get the feeling of "Where the * did my codelists go" haha 


=> it is ok that we show the archived codelists in the overview the only thing is that the user cannot be able to edit or update it.  Alert plus only readonly data

Will update the test steps thanks for noticing

ZijneDwaasheid:
> @ZijneDwaasheid
> 
> > Ok so tested this and followed the 'how to test', at step 3 You may not be able to see the codelists in the overview. So first of all it gets the tag archived, but then you see it of course still in the overview and you're still able to add the concept-scheme in the formbuilder. I don't know this what you expected or I'm missing something here.
> 
> Ah im sorry didn't update the test description but this is indeed not the expected behaviour that was written down. With the thing in mind how should we recover from an archived list I didn't remove it from the overview page and just added the tag. This makes it possible for the user to see the codelist so they not get the feeling of "Where the * did my codelists go" haha
> 
> => it is ok that we show the archived codelists in the overview the only thing is that the user cannot be able to do is editing. Alert plus only readonly data
> 
> Will update the test steps thanks for noticing

alright, that's what i thought, just making sure.

translations/messages/nl-be.yaml
```diff
@@ -5,6 +5,7 @@ messages:
     characters: "Karakters"
     deleted: "Verwijderd"
     upToDate: "Up-to-date"
+    archived: "Ge-archiveerd"
```
ZijneDwaasheid, line 8: One small thing, I think it's just 'gearchiveerd'.

ZijneDwaasheid (approved):

translations/messages/nl-be.yaml
JonasVanHoof: Haha you are right!

app/components/concept-scheme-uri-selector.js
```diff
@@ -56,7 +56,11 @@ export default class ConceptSchemeUriSelectorComponent extends Component {
       },
     });
 
-    this.options = sortObjectsOnProperty([...conceptSchemes], 'label');
+    const notArchivedConceptSchemes = [...conceptSchemes].filter(
+      (scheme) => !scheme.isArchived
+    );
+
+    this.options = sortObjectsOnProperty(notArchivedConceptSchemes, 'label');
```
JonasVanHoof, line 63: @ZijneDwaasheid just thought about this that the archived conceptschemes will still be shown in the concept scheme selector even if they are archived.. 

Added this filter method to only get private and public conceptschemes that are not archived 👌🏻  when a form is created with an archived conceptscheme it will still show in the form preview and test as it is still in the database :)

Sorry about this.. Could you check it again :)

JonasVanHoof: Maybe test not yet as it is not using the new user-test page yet 🙈

ZijneDwaasheid (approved):
For other ticket, analyse how we could let the user know that the concept-scheme that they are using is a archived concept-scheme

PR #121, opened 2024-02-15
SFB-165: concept scheme editor go back to form button

ZijneDwaasheid:
## ID
SFB-165

 ## Description
Instead of going back to the index page of the formbuilder, we now have the option to go back to form where we clicked the button 'codelijsten'. If we landed on the codelijsten page on a different way, not from a form, the return button will bring you to the index of the formbuilder.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Try to do something that causes something that is not expected, like described in the description of this pr.


app/services/formbuilder-id-service.js
```diff
@@ -0,0 +1,19 @@
...
+  setFormbuilderId(id) {
+    this.formbuilderId = id;
+  }
+
+  getFormbuilderId() {
+    return this.formbuilderId;
+  }
+
+  clearFormbuilderId() {
+    this.formbuilderId = EMPTY_FORMBUILDER_ID;
```
JonasVanHoof, line 17: Would just set it to null

app/services/formbuilder-id-service.js
```diff
@@ -0,0 +1,19 @@
+import Service from '@ember/service';
+
+const EMPTY_FORMBUILDER_ID = '';
```
JonasVanHoof, line 3: Can be removed than as set to null in clear method

app/services/formbuilder-id-service.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default class FormbuilderIdServiceService extends Service {
```
JonasVanHoof, line 5: Not a fan of the service name tbh..

* formbuilder => the application
* id => managing the application instance id's? (blowing things up 😁 )
* service is twice in the name

Did you check the sticky router params that @aatauil mentioned?

A service is a little to big for just going back one route in one place IMO if the sticky query router param does not work. Not sure if this is something but we can maybe store it in the storage of the browser? @aatauil

JonasVanHoof (commented):
It works but some remarks that should be checked first

app/services/formbuilder-id-service.js
ZijneDwaasheid: yes will check on that

JonasVanHoof:
@ZijneDwaasheid  this one can be closed no?

ZijneDwaasheid:
> @ZijneDwaasheid this one can be closed no?

yes indeed

PR #125, opened 2024-02-19
SFB-165: using query parameters to pass the form id to codelijsten index

ZijneDwaasheid:
## ID
SFB-165

 ## Description

using the a service to pass and hold the id was not ideal so doing it the right way now

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

 ## Links to other PR's

 - https://github.com/lblod/frontend-form-builder/pull/121

### DIFF
--- a/app/controllers/codelijsten/index.js
+++ b/app/controllers/codelijsten/index.js
@@ -1,10 +1,21 @@
 import Controller from '@ember/controller';
 import { service } from '@ember/service';
+import { action } from '@ember/object';
 
 export default class CodelijstenController extends Controller {
   @service features;
+  @service router;
 
   sort = '-preflabel';
   page = 0;
   size = 20;
+
+  @action
+  returnToFormBuilder() {
+    if (this.formId) {
+      this.router.transitionTo('formbuilder.edit', this.formId);
+    } else {
+      this.router.transitionTo('index');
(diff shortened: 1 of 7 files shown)


JonasVanHoof (changes requested):
This url will throw an error and the user will stay on the same page. http://localhost:4200/codelijsten?formId=65D34FFCD9283D870DF72DB6ABCCC

When not intentially breaking it lke above it works good, thanks!

app/controllers/codelijsten/index.js
```diff
@@ -1,10 +1,21 @@
...
 export default class CodelijstenController extends Controller {
   @service features;
+  @service router;
 
   sort = '-preflabel';
   page = 0;
   size = 20;
+
+  @action
+  returnToFormBuilder() {
```
JonasVanHoof, line 14: Would name it a bit more specific. "returntoFormBuilder()" and the next line will add an id in the route. Maybe just go with "goToFormWithId" or "routeToOpenForm" or "routeToLastOpenedForm"

JonasVanHoof (approved):

JonasVanHoof (approved):

PR #126, opened 2024-02-19
SFB-178: Rich Text Editor

JonasVanHoof:
## ID
 [SFB-178](https://binnenland.atlassian.net/browse/SFB-178)

 ## Description

Create a rich text editor so the user can add basic styling to their text.
 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Use the rich text editor functionality in the form builder

 ## Links to other PR's

 - /


JonasVanHoof:
The editor looks like this atm. I removed some functionality based on what comes through in the preview of the RDF form. As you see that the italic text is not italic in the previewI will also remove it after this comment.

image

JonasVanHoof:
> Checked with Marie and we will be adding these plugins for the editor. [ref to Jira](https://binnenland.atlassian.net/browse/SFB-178?focusedCommentId=98066)
> 
> * text style (bold, italic, underline, ~strike-through, subscript, superscript~)
> * headings
> * highlight
> * list
> * link (not sure about this one, I think it would depend on the form and field, for now I would leave it out)

It looks like this now
image

JonasVanHoof:
As just discussed on standup we will not be using the richTextEditor displayType in our builder. The user is able to add it to there form but this is only experimental.It will not show up in the inputType option dropdown but the smart user could just type it in the ttlCode.

This is the example that the user can view and use the richTextEditor on the test page of the form
image

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -0,0 +1,172 @@
...
+  ordered_list,
+} from '@lblod/ember-rdfa-editor/plugins/list';
+import {
+  bullet_list_input_rule,
+  ordered_list_input_rule,
+} from '@lblod/ember-rdfa-editor/plugins/list/input_rules';
+import { placeholder } from '@lblod/ember-rdfa-editor/plugins/placeholder';
+import { highlight } from '@lblod/ember-rdfa-editor/plugins/highlight/marks/highlight';
+import { color } from '@lblod/ember-rdfa-editor/plugins/color/marks/color';
+import { inputRules } from '@lblod/ember-rdfa-editor';
```
JonasVanHoof, line 48: Imported all of these as when removing some certain plugins stopped working correctly.

app/components/rdf-form-fields/rich-text-editor.js
```diff
@@ -0,0 +1,172 @@
...
+
+export default class RichTextEditorComponent extends SimpleInputFieldComponent {
+  @tracked editorController;
+  inputId = 'richtext-' + guidFor(this);
+
+  defaults = {
+    highlightColor: '#FFEA00', // Yellow
+    textColor: '#000000', // Black
```
JonasVanHoof, line 56: think this is ok to have as defaults. We could add the colors to our translation file but  that is not needed atm IMO

app/styles/app.scss
```diff
@@ -4,10 +4,13 @@
 @import "project/c-form-builder-edit";
 @import "project/c-form-test";
 @import "project/c-app-toolbar";
+@import "project/c-rich-text-editor";
 @import "project/_c-app-codelijsten";
 
 @import "shame";
 
+@import "ember-rdfa-editor";
```
JonasVanHoof, line 12: Import all the stylings of the RDFa editor so it looks like pretty

app/templates/formbuilder/edit.hbs
```diff
@@ -96,7 +96,7 @@
           <AuLoader />
         {{else}}
           <RdfForm
-            @groupClass="au-u-margin-left-small au-u-margin-top"
+            @groupClass="au-u-margin-left au-u-margin-right au-u-margin-top"
```
JonasVanHoof, line 99: Added for the scrollbar not to be in the way of the text of the form

public/forms/builder/meta.ttl
```diff
@@ -71,6 +71,12 @@ display:textArea
   core:uuid "2e114946-3f3e-43ee-af16-d7c7255977e3" ;
   skos:inScheme <http://lblod.data.gift/concept-schemes/c5a91bd7-3eb5-4d69-a51b-9bac6bf345f6> ;
   skos:prefLabel  "Text area"@en .
+display:richTextEditor
+  rdf:type  rdfs:Class , skos:Concept ;
+  ext:displayType "richTextEditor" ;
+  ext:usesConceptScheme "false"^^xsd:boolean ;
+  # skos:inScheme <http://lblod.data.gift/concept-schemes/c5a91bd7-3eb5-4d69-a51b-9bac6bf345f6> ; This is commented as it is experimental at this moment of time
```
JonasVanHoof, line 78: As comment says. Experimental. Can be used directly in the ttl but the builder will not show the option

JonasVanHoof (commented):
@ZijneDwaasheid 

Add some comments to make clear why some things are done like that, hope it helps with understanding :)

ZijneDwaasheid (commented):
It does what it's supposed to do. Looks good to me!
(again for the future, we also need to think about validations)

JonasVanHoof:
> It does what it's supposed to do. Looks good to me! (again for the future, we also need to think about validations)

Yes indeed, didn't add validations to the field type yet but i can see that they would like to add a required validation to it. Max characters will be a bit harder as the html tags will be included in the character count.

Maybe create an analyse ticket for it, put it in the backlog at the top so we will notice it + mention the ticket here.

PR #131, opened 2024-02-28
SFB-196: add search input field for the form overview

JonasVanHoof:
## ID
 [SFB-196](https://binnenland.atlassian.net/browse/SFB-196)

 ## Description

As a user I want to search my form by name. this is done by using the `mu-cl resource` filters.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. go to the form overview
2. a search input should be visible
3. when typing in the search input the data will update to the requested values

 ## Links to other PR's

 - /


ZijneDwaasheid (approved):
maybe also take a look why it also searches on description, the problem here is that the description isn't shown in the frontend anymore, in the backend it still exists. So when searching on a certain name you also get the ones with that name in description.

aatauil (approved):

PR #133, opened 2024-02-29
SFB-199: new builder SKELETON

JonasVanHoof:
## ID
 [SFB-199](https://binnenland.atlassian.net/browse/SFB-199)

 ## Description

Create the skeleton on how the builder will look like. This means a sidebar and the segments we want to create are visible when going to route `/formbuilder/:id/editor`

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Click on a form that you want to edit
2. See if looks like the wire frame in the attachments

 ## Links to other PR's

 - /

## Attachments

image

### DIFF
--- a/app/components/editor/building-view.hbs
+++ b/app/components/editor/building-view.hbs
@@ -1,0 +1,1 @@
+<h1>building view</h1>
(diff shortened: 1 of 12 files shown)


JonasVanHoof:
Not focusing on this atm

PR #141, opened 2024-03-07
SFB-211: table configuration

JonasVanHoof:
## ID
 [SFB-211](https://binnenland.atlassian.net/browse/SFB-211)

 ## Description

 As a user I want to change the configuration of a table listing

1. minCount
2. maxCount
3.  form:tableIndexLabel 
4. form:showTableRowIndex

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Update the configuration of a table

 ## Links to other PR's

 - /

PR #142, opened 2024-03-08
SFB- 212/table config card for multiple tables in section

JonasVanHoof:
## ID
 [SFB-212](https://binnenland.atlassian.net/browse/SFB-212)

 ## Description

We want to show the config table card to show for every tbale that is in the selected section.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

Will see multiple table configs for every table in the section

 ## Links to other PR's

 - https://github.com/lblod/frontend-form-builder/pull/141

### DIFF
--- a/app/components/concept-scheme-uri-selector.hbs
+++ b/app/components/concept-scheme-uri-selector.hbs
@@ -4,6 +4,8 @@
     @options={{this.options}}
     @selected={{this.selected}}
     @onChange={{this.setSelected}}
+    @searchEnabled={{true}}
+    @searchField="label"
     as |conceptScheme|
   >
     {{conceptScheme.label}}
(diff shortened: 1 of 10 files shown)


ZijneDwaasheid (approved):
looks good, maybe something small. You can give a higher min value than max value. But that's kind of ok I think because you have a fixed table where you cant add or remove rows

JonasVanHoof:
> looks good, maybe something small. You can give a higher min value than max value. But that's kind of ok I think because then you have a fixed table where you cant add or remove rows

Hm good one. Would it be logic to implement in our application or would it be something that the table must handle?

PR #144, opened 2024-03-12
SFB-217: rename aggregate field to begrotingstabelTotal component

JonasVanHoof:
## ID
 [SFB-217](https://binnenland.atlassian.net/browse/SFB-217)

 ## Description

This is the start for creating a component that will calculate and compare the values in two columns from two separate tables. This result will be shown in this new component.  Changes to this component will be done in another PR and will be linked at the bottom in _Links to other PR's_

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

1. Can use the aggregate field as before only need to rename the displayType to `begrotingstabelTotal`

 ## Links to other PR's

 - https://github.com/lblod/frontend-form-builder/pull/146

PR #146, opened 2024-03-12
SFB-218: logic begrotingstabel total field

JonasVanHoof:
## ID
 [SFB-218](https://binnenland.atlassian.net/browse/SFB-218)

 ## Description

We want to calculate the sum of values in a column. These values will be shown in a total field. There will be added more fields depending on how many targets you add to the begrotingsTotal field. These values will be shown in the form preview. When these values do not end up being equal to each other we will show a message that it is not valid.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Create a new form from template. Choose `Begrotingstabel`
2. Start adding values to the `Inkomsten` and `Uitgaven` tables
3. If the result values do not match each other a error message is shown

 ## Links to other PR's

 - https://github.com/lblod/frontend-form-builder/pull/144

 ## Attachements

**V1**

image

**V2**

image

**V3**

image

### DIFF
--- a/app/components/rdf-form-fields/begrotingstabel-total-field.js
+++ b/app/components/rdf-form-fields/begrotingstabel-total-field.js
@@ -42,7 +42,7 @@ export default class BegrotingstabelTotalFieldComponent extends Component {
       .map((st) => st.object);
 
     if (collections.length !== 2) {
-      this.collectionErrorMessage = `Tabel wordt niet geupdate omdat er geen twee targets gedifneerd zijn naar de waarde`;
+      this.collectionErrorMessage = `Tabel heeft te weing informatie. Voeg twee xsd:targets toe. bv. xsd:target ( ext:Expense ext:amount ), ( ext:Income ext:amount )`;
       return;
     }
     this.collectionErrorMessage = null;
@@ -110,7 +110,7 @@ export default class BegrotingstabelTotalFieldComponent extends Component {
         (item) => item.calculationResult == this.totals[0].calculationResult
       )
     ) {
-      this.comparisonWarningMessage = `Results are not the same`;
+      this.comparisonWarningMessage = `De resultaten zijn niet gelijk`;
 
       return;
     }
--- a/app/components/rdf-form-fields/begrotingstabel-total-field.js
+++ b/app/components/rdf-form-fields/begrotingstabel-total-field.js
@@ -36,26 +36,26 @@ export default class BegrotingstabelTotalFieldComponent extends Component {
   }
 
   calculateTotals = restartableTask(async () => {
-    const target = this.store.any(
-      this.field.uri,
-      XSD('target'),
-      undefined,
-      this.graphs.formGraph
-    );
+    const collections = this.store
+      .match(this.field.uri, XSD('target'), undefined, this.graphs.formGraph)
+      .map((st) => st.object);
 
-    const options = {
-      path: target,
-      store: this.store,
-      formGraph: this.graphs.formGraph,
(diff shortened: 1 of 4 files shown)


app/components/rdf-form-fields/begrotingstabel-total-field.js
```diff
@@ -0,0 +1,168 @@
...
+    }).format(amount);
+  }
+
+  calculateTotals = restartableTask(async () => {
+    const collections = this.store
+      .match(this.field.uri, XSD('target'), undefined, this.graphs.formGraph)
+      .map((st) => st.object);
+
+    if (collections.length !== 2) {
+      this.collectionErrorMessage = `Tabel heeft te weing informatie. Voeg twee xsd:targets toe. bv. xsd:target ( ext:Expense ext:amount ), ( ext:Income ext:amount );`;
```
JonasVanHoof, line 45: Should we add a translation file for just the `BegrotingstabelTotal` component?

app/components/rdf-form-fields/begrotingstabel-total-field.js
```diff
@@ -0,0 +1,168 @@
...
+    this.setWarningMessage();
+  });
+
+  getIndexOfTotal(collectionId) {
+    return this.totals.findIndex((item) => item.id == collectionId);
+  }
+
+  isValidIndex(index) {
+    return index == 0 || index !== -1;
+  }
```
JonasVanHoof, line 94: Generic utils method? is used in two places now

app/components/rdf-form-fields/begrotingstabel-total-field.js
```diff
@@ -0,0 +1,168 @@
...
+
+    if (
+      !this.totals.every(
+        (item) => item.calculationResult == this.totals[0].calculationResult
+      )
+    ) {
+      this.comparisonWarningMessage = `De resultaten zijn niet gelijk`;
+
+      return;
+    }
```
JonasVanHoof, line 116: Could also check if value on index 0 and 1 are the same

app/components/rdf-form-fields/begrotingstabel-total-field.js
JonasVanHoof: These translations can than be easely added to the project that will be using this component.

ZijneDwaasheid (approved):

public/forms/templates/begrotingstabel.ttl
```diff
@@ -0,0 +1,140 @@
...
+        nodes:f0d14635-879d-44ff-ae9b-2c34ed278f8a;
+    form:partOf
+        nodes:4ce76f84-e71d-4d2b-8733-7fb02cbb12f7,
+        nodes:a4503b20-7c8f-497c-abac-9650691cec75,
+        nodes:f9dee3f6-e782-417f-ad89-c54e603a6a04 .
+
+ext:expenseGenerator
+    a form:Generator;
+    form:dataGenerator form:addMuUuid;
+    form:prototype [ form:shape [ a ext:Expense; ext:amount 0.0 ] ].
```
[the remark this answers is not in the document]
JonasVanHoof: It depends, when clicking before the zero and typing your number the zero dissapears but if you click after the zero it means you want to make the number bigger.

app/components/rdf-form-fields/begrotingstabel-total-field.js
JonasVanHoof: No as this is a component that needs to be copied to other projects aswell if they want to use it there

PR #147, opened 2024-03-12
SFB-214/Make custom validation message optional

ZijneDwaasheid:
## ID
SFB-214

 ## Description

The default message is automatically added when adding a validation to field, if you want to change the message you can check a box and a input field will appear. 

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [x] Other

 ## How to test

Add validation and check messages.

### DIFF
--- a/app/components/rdf-form-fields/error-message-input-field.hbs
+++ b/app/components/rdf-form-fields/error-message-input-field.hbs
@@ -4,14 +4,18 @@
     {{@field.label}}
 </AuLabel>
 
-<AuInput
-  @error={{this.hasErrors}}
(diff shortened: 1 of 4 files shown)


JonasVanHoof (approved):
Works.
Maybe create a ticket with in mind to disabled the dropdown when a validation is selected so the user is forced to delete the validation when wanting to change it. This would make that our component is always updated with the latest ttl code configuration

JonasVanHoof:
Maybe also an option to add the section name above all the fields of that section. I saw that you can add groups to the dropdown
https://ember-power-select.com/docs/the-list

PR #150, opened 2024-03-15
SFB-226: Remove query with size 999

JonasVanHoof:
## ID
 [SFB-226](https://binnenland.atlassian.net/browse/SFB-226)

 ## Description

When going to the codelists page it returns a 500. The reason is why is unsure yet.

As from the beginning of the concept scheme selector a call was done towards the database to get all the concept schemes in the database. This call was done by adding a parameter `page: { size: 9999 }`. This is not great as if ever the list becomes larger than 9999 it will not show the latest conceptschemes. I updated the call so all the items are get per page. This will make that the amount of calls will increase to get the information, everything will always be get from the database and there is no call anymore that uses a constant of 9999 to get all the items in the store. It does not feel the correct way but it's already better than the implementation of today. If someone else comes up with a better way please just tell me a way.

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

1. add a dropdown to the form
2. Go to the configuration tab and select a codelist
3. All of the codelists should be available

 ## Links to other PR's

 - /

### DIFF
--- a/app/components/concept-scheme-uri-selector.hbs
+++ b/app/components/concept-scheme-uri-selector.hbs
@@ -4,6 +4,8 @@
     @options={{this.options}}
     @selected={{this.selected}}
     @onChange={{this.setSelected}}
+    @searchEnabled={{true}}
+    @searchField="label"
     as |conceptScheme|
   >
     {{conceptScheme.label}}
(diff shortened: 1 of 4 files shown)


JonasVanHoof:
@ZijneDwaasheid  something we would like to do in our code? Really did not like the constant of 999. Ths would do more smaller calls what isn't actually a bad thing

PR #152, opened 2024-03-19
SFB-231: fix shacl warning on resultMessage

ZijneDwaasheid:
## ID
SFB-231

 ## Description

When adding a resultMessage to a validation it gave a wrong warning.

 ## Type of change

 - [x] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

If you add 1 string or langString it shouldn't give a warning. If you add more than 1 or the wrong datatype it will give a warning.


JonasVanHoof (approved):
Looks good

PR #155, opened 2024-03-21
SFB-234: add exact value constraint shacl shape

JonasVanHoof:
## ID
 [SFB-234](https://binnenland.atlassian.net/browse/SFB-234)

 ## Description

A shape for validaiton "Exact value constraint" was not yet added to the list off shapes. With this PR we add this shape to the list so next time someone uses this validation wrong we get feedback on that.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. run the test of the exact value constraint
2. Add the validation to your field and try to make break it  by filling in data that is not supposed to be there

 ## Links to other PR's

 - /


ZijneDwaasheid (approved):
Maybe we also want to check on the datatype of the custom value, or doesn't it matter?

JonasVanHoof:
> Maybe we also want to check on the datatype of the custom value, or doesn't it matter?

Not sure if this is needed. Can you break it with a string?

ZijneDwaasheid:
> > Maybe we also want to check on the datatype of the custom value, or doesn't it matter?
> 
> Not sure if this is needed. Can you break it with a string?

No it works with a string and a number

PR #156, opened 2024-03-21
SFB-235: add shape + test for valid phone number constraint

JonasVanHoof:
## ID
 [SFB-235](https://binnenland.atlassian.net/browse/SFB-235)

 ## Description

A shape for validation "Valid phone number" was not yet added to the list off shapes. With this PR we add this shape to the list so next time someone uses this validation wrong we get feedback on that.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. run the test of the exact value constraint
2. Add the validation to your field and try to make break it  by filling in data that is not supposed to be there

 ## Links to other PR's

 - /

### DIFF
--- a/public/SHACL/V2/valid-phone-number-constraint.ttl
+++ b/public/SHACL/V2/valid-phone-number-constraint.ttl
@@ -1,0 +1,50 @@
+#####################################################
+# Valid phone number constraint shape for V2 forms ##
+#####################################################
+
+form:ValidPhoneNumber
+  a sh:NodeShape ;
+  sh:targetClass form:ValidPhoneNumber ;
+  sh:property
+    [
+      sh:path form:grouping ;
+      sh:hasValue form:MatchEvery ;
+      sh:minCount 1 ;
+      sh:maxCount 1 ;
+      sh:message "form:Grouping moet de waarde form:MatchEvery zijn."@nl ;
+      sh:severity sh:Warning;
+    ],
+    [
+      sh:path sh:path ;
+      sh:minCount 1 ;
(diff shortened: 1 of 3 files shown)


ZijneDwaasheid (commented):
I would split up the min max count from the sh:in, so we can give it a different message. So when accidentally added two it gives the right message

public/SHACL/V2/valid-phone-number-constraint.ttl
```diff
@@ -0,0 +1,50 @@
...
+      sh:severity sh:Warning;
+    ],
+    [
+      sh:path form:defaultCountry ;
+      sh:datatype xsd:string ;
+      sh:minCount 1 ;
+      sh:maxCount 1 ;
+      sh:in (
+        "BE"
+        "NL"
```
ZijneDwaasheid, line 30: over here

JonasVanHoof:
> I would split up the min max count from the sh:in, so we can give it a different message. So when accidentally added two it gives the right message

Your comments are very hard to decipher. Add the comments to a line of code please ah you just did XD thanks!

public/SHACL/V2/valid-phone-number-constraint.ttl
JonasVanHoof: In who's mind is adding two country codes to a phonenumber a thing  XD

I'll add it

ZijneDwaasheid: yeah mistakes are often not intentionally

ZijneDwaasheid (approved):

PR #157, opened 2024-03-21
SFB-242: unsaved changed modal when leaving codelijsten.edit

JonasVanHoof:
## ID
 [SFB-242](https://binnenland.atlassian.net/browse/SFB-242)

 ## Description

When routing a away from the `codelijsten.edit` route and the user has unsaved changes the user now has to confirm if they want to save or discard the changes.

The original data is set back to the current view when pressing discard. When pressing the save button it does the same as the current save button.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

1. Update a codelist and route back with the arrow in the breadcrumbar
2. A modal will be shown as you did not save your latest changes
3. try out both options

 ## Links to other PR's

 - /

### DIFF
--- a/app/controllers/codelijsten/edit.js
+++ b/app/controllers/codelijsten/edit.js
@@ -338,21 +338,24 @@ export default class CodelijstenEditController extends Controller {
 
     if (
       this.isValidConceptSchemeName() &&
-      this.isValidConceptSchemeDescription() &&
-      this.isConceptListIncludingEmptyValues()
+      this.isValidConceptSchemeDescription()
     ) {
       if (
         this.isCodelistDescriptionDeviating() ||
-        this.isCodelistNameDeviating() ||
-        this.isConceptListChanged() ||
-        this.conceptsToDelete.length >= 1
+        this.isCodelistNameDeviating()
       ) {
         this.isSaveDisabled = false;
       } else {
         this.isSaveDisabled = true;
       }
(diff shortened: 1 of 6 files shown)


app/controllers/codelijsten/edit.js
```diff
@@ -322,33 +330,39 @@ export default class CodelijstenEditController extends Controller {
...
-          this.conceptsToDelete.length >= 1
-        ) {
-          this.isSaveDisabled = false;
-        } else {
-          this.isSaveDisabled = true;
-        }
-      } else {
-        this.isSaveDisabled = true;
-      }
-    } else {
```
JonasVanHoof, line 349: Hopefully I cleared this up with the rewrite? 😃

JonasVanHoof:
I think I handled everything, could you check 😃

ZijneDwaasheid (commented):
looks ok but I found something. When you save a new concept-scheme, then revisit the same concept-scheme and edit it again you can delete the description and save it without the description (only thing you need to do is add a concept). You can also do it with the name, but then weird things happen and it kind of breaks. I think you have to look at how it reroutes after it deletes something.

ZijneDwaasheid:
@JonasVanHoof If you need more info on what I wrote, just send me something on rocketchat

PR #159, opened 2024-03-26
SFB-229: valdiation tab with custom components

JonasVanHoof:
## ID
 [SFB-229](https://binnenland.atlassian.net/browse/SFB-229)

 ## Description

 The validation tab is not created by generating a form. This form is than merged with the current form so the validations are added. 

In this PR we will create custom components to show these validations in the formbuilder. The view should be almost the same and the result should be the same. The difference is that some things will work more smoothly as we are not limited with the generation of forms.

 ## Type of change

 - [ ] Bug fix
 - [x] New feature
 - [x] Breaking change
 - [ ] Other

 ## How to test

Add remove and update validations on fields

 ## Links to other PR's

 - /


JonasVanHoof:
It's a big PR but everything should be rather easy to read and understand 🆗  @ZijneDwaasheid

app/components/validation/card.js
```diff
@@ -0,0 +1,229 @@
...
+    if (this.validationType) {
+      const messageLiteral = this.metaStore.any(
+        this.validationType.object,
+        SHACL('resultMessage'),
+        undefined,
+        this.metaGraph
+      );
+
+      if (messageLiteral) {
+        message = messageLiteral.value;
```
ZijneDwaasheid, line 71: Not that important I guess, but the language tag will not be used

app/components/validation/card.hbs
```diff
@@ -0,0 +1,57 @@
...
+      </c.content>
+      <c.footer>
+        <div class="au-u-flex au-u-flex--end au-u-margin-top-small">
+          <AuButton
+            @disabled={{this.isUpdatingValidations}}
+            @skin="secondary"
+            @alert={{true}}
+            {{on "click" this.deleteValidation}}
+          >
+            {{this.deleteButtonText}}</AuButton>
```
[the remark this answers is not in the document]
JonasVanHoof: I think it just looks better :)
We are already using two types of delete buttons. In the builder the first one in the screenshot is being used. When we go to the save un saved changes modal we are using the first one in the screenshot.

Now I do't think we have a specific styling defined for the delete buttons but I could also go with the style of the delete button in the builder. I could live with that :)  What do you think?

image

app/components/validation/type-selector.js
```diff
@@ -0,0 +1,37 @@
...
+        this.selectedType = validationOption;
+        this.isDisabled = true;
+      }
+    }
+  }
+
+  @action
+  updatedSelectedValidation(newValidationType) {
+    this.selectedType = newValidationType;
+    this.isDisabled = true;
```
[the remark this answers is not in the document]
JonasVanHoof: I really like the flow now when it is disabled. It shows that you have added the selected type and that you have to stick with it or remove the validation. Next to that the actions made in the ttl are very clear/small what makes it less maintenance and debug time if something changes

app/components/validation/card.js
```diff
@@ -0,0 +1,229 @@
...
+        return { subject: t.subject, label: label && label.value };
+      });
+
+    const optionsForFieldType = allOptions.filter((option) => {
+      return conceptOptions
+        .map((concept) => concept.value)
+        .includes(option.subject.value);
+    });
+
+    const itemsWithDisabledOptions = optionsForFieldType.map((option) => {
```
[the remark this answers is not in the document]
JonasVanHoof: Adding a new empty one when a duplicate is found does not feel like the way to go tbh. if you have many duplicates it will add all empty validations. 

Achraf said that we indeed could just delete the duplicate validations but  that's not making the user aware of that this happend. When a bug is reported because validations are missing  I don't want to be the one that needs to find the if statement that checks if duplicate and than forcefully remove it.

We'll check with @aatauil

app/components/validation/card.hbs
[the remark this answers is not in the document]
ZijneDwaasheid: I see, yeah this will all be checked by a ux/ui designer anyway. For now it's good

app/components/validation/card.js
ZijneDwaasheid: yeah but does it hurt to add it. I mean it is given in the meta.ttl

app/components/validation/card.js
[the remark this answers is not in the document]
JonasVanHoof: Achraf said that it is good for now. Its _acceptable_

app/components/validation/card.js
JonasVanHoof: What would you be adding? I can think of a selector component that will show the options `@nl` and `@en` but would you use it and is it really an addition?

JonasVanHoof: Ah I see sorry, the default message that is being added has no language tag. They are all @nl I would leave it. Like I said above if users change the result message it is not added and when they don't there is only one message that can be used

ZijneDwaasheid: I was thinking only in case of default message, as it is given this way. Giving an extra selector for the user is maybe overkill.

JonasVanHoof: If you like you can create an analyse ticket for this and link it to SFB-229 🙂

JonasVanHoof:
No more bugs or missing functionality?

PR #160, opened 2024-03-27
SFB 274: add syntax error message to cosnole

ZijneDwaasheid:
## ID
SFB 274

 ## Description
After we added the shacl shape validation messages to the console, we want also the syntax error messages to be in the console.

 ## Type of change

 - [ ] Bug fix
 - [ ] New feature
 - [ ] Breaking change
 - [ ] Other

 ## How to test

 Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

 ## Links to other PR's

### DIFF
--- a/app/controllers/formbuilder/edit/code.js
+++ b/app/controllers/formbuilder/edit/code.js
@@ -105,7 +105,12 @@ export default class FormbuilderEditCodeController extends Controller {
         'text/turtle'
       );
     } catch (error) {
-      console.warn({ caught: error });
+      // console.warn({ caught: error });
+      this.consoleMessages.pushObject({
+        severity: this.getConsoleSeverity(error.severity),
+        subject: error,
+        
+      });
       // This is limiting the errors thrown in the console while editing the code
       return;
     }


app/templates/formbuilder/edit/code.hbs
```diff
@@ -17,6 +17,12 @@
       >
         {{t "codeConsole.buttonNotification"}}:
         {{this.consoleMessages.length}}
+        {{#if this.ttlHasErrors}}
+          <AuPill @skin="error" @icon="cross">Error</AuPill>
+        {{/if}}
+        {{#if this.ttlHasWarnings}}
+          <AuPill @skin="warning" @icon="cross">Warning</AuPill>
+        {{/if}}
       </AuButton>
```
JonasVanHoof, line 26: Adding pills in a button?

Could you move the pills to the right side of the console bar? 
image

I would do or text or the icon in the pill but not both

JonasVanHoof (commented):
I think we should check on how we can have multiple messages in the console. When removing the paths from a validation and than causing a syntax error the warning of the path disappears and the messages are reset to a list with that single error. It's possible to keep all the previous messages when we are not clearing the messages when we validate. 

Give it some more thought this can be done better :)

